PR #7, opened 2019-04-26
If RecyclerView at offset 0, scroll to offset 0 after inserting a new item at 0.

### LINKED ISSUE #6: When inserting an item at position 0 after the list is rendered, need to manually scroll up to see the item
It is really a RecyclerView issue: when we insert an item in a RecyclerView at position 0, the RecyclerView does not scroll up or down.
This is of course wanted behaviour in most cases, except for one: when you are at scroll offset 0, it would be better to stay at scroll offset 0 and display the new item.

The use case that I have is the following. You are on the home page of your app. And you query some asynchronous service to query for some notifications banners.
It's like, "hot informations" that are displayed on top of the page.
Your asynchronous service returns info, you create a banner, and insert it at index 0. You will actually need to manually scroll up to see it.

redwarp:
When the RecyclerView is scrolled all the way to the top, inserting an item at index 0 will show the item, instead of not scrolling.
This change also adds a menu item in the sample app to showcase this new behaviour.

Closes #6

### DIFF
--- a/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
+++ b/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
@@ -131,7 +131,7 @@ public class RecyclerViewComponentController implements ComponentController {
         mRecyclerView.setLayoutManager(mLayoutManager);
         mRecycledViewPool = mRecyclerView.getRecycledViewPool();
         setupComponentSpans();
-        addVisibilityListener();
+        addVisibilityListeners();
     }
 
     @Override
@@ -243,10 +243,8 @@ public class RecyclerViewComponentController implements ComponentController {
     @Override
     public void clear() {
         mComponentGroup.clear();
-        mRecyclerView.removeOnScrollListener(mOnScrollListener);
-        mComponentGroup.unregisterComponentDataObserver(mComponentVisibilityListener);
-        mRecyclerViewAdapter.unregisterAdapterDataObserver(mAdapterDataObserver);
-        addVisibilityListener();
+        removeVisibilityListeners();
+        addVisibilityListeners();
     }
 
     @Override
@@ -270,7 +268,7 @@ public class RecyclerViewComponentController implements ComponentController {
         }
     }
 
-    private void addVisibilityListener() {
+    private void addVisibilityListeners() {
         mComponentVisibilityListener =
                 new ComponentVisibilityListener(
                         new RecyclerViewLayoutManagerHelper(mLayoutManager), mComponentGroup);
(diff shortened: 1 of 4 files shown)


bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -59,13 +60,11 @@ public RecyclerViewComponentController(@NonNull RecyclerView recyclerView) {
...
      * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
      * the {@link RecyclerView}. In order to make the lanes (columns / rows), this component
-     * controller will set the {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do 
+     * controller will set the {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do
      * it manually.
      */
     public RecyclerViewComponentController(
-            @NonNull RecyclerView recyclerView,
-            @Orientation int orientation
-    ) {
```
argo49, line 68: Just curious, are you using a kotlin formatter? Is it something we should consider adding to bento?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -284,8 +286,29 @@ public void onScrolled(RecyclerView recyclerView, int dx, int dy) {
                         mComponentVisibilityListener.onScrolled();
                     }
                 };
+        mAdapterDataObserver =
```
argo49, line 287: nit: Maybe change this method to `addVisibilityListeners` since we're doing more than one thing here.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -244,6 +245,7 @@ public void clear() {
         mComponentGroup.clear();
         mRecyclerView.removeOnScrollListener(mOnScrollListener);
         mComponentGroup.unregisterComponentDataObserver(mComponentVisibilityListener);
+        mRecyclerViewAdapter.unregisterAdapterDataObserver(mAdapterDataObserver);
```
argo49: nit: Maybe we should group these in a private method `removeVisibilityListeners()` so that it's more obvious that there's a symmetry where we want to eventually remove whatever we add.

argo49 (changes requested):

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
redwarp: No, but I am using a Java formatter: https://github.com/google/google-java-format

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
redwarp: Makes sense!

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
redwarp: Makes sense as well.

redwarp:
Updated, please check changes again

argo49 (approved):

argo49:
@redwarp I think the only thing that's missing is a minor version bump

PR #18, opened 2019-06-13
Enabled drag and drop for Bento.

wyattjsmith1:
This is a preliminary commit for drag and drop in Bento. I am just looking to see if I am headed in a good direction for architecture, then I will flush out all of the details and make it production ready.

For testing, you can run the sample app and click on `Reorder Items`. The page is ugly, but works to demo the feature. There are three sections. One that can not be moved (normal). One that can be rearranged by long pressing and then dragging. Lastly, there is a particularly ugly section for dragging by the handle.

One known limitation of this is that you can long press outside of a handle and still move an item with a handle. I can't seem to find a way to avoid this, but I don't think it's a big deal because you can reorder it anyways.

Another thing to consider: we don't want users to move items to anywhere on the page, so items can only be moved within their component. That means you can not drag an item with a handle to the group with the long press. This means that developers will be required to encompass their entire rearrangeable list within one `Component`. I think that is fine, but I am curious if there are other opinions on this.

### DIFF
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
@@ -9,6 +9,7 @@ import androidx.appcompat.app.AppCompatActivity
 import com.yelp.android.bento.componentcontrollers.RecyclerViewComponentController
 import com.yelp.android.bento.components.ListComponent
 import com.yelp.android.bento.components.OnItemMovedCallback
+import com.yelp.android.bento.core.Component
 import com.yelp.android.bento.core.ComponentGroup
 import com.yelp.android.bento.core.ComponentViewHolder
 import com.yelp.android.bento.core.setOnDragStartListener
@@ -32,6 +33,7 @@ class ReorderListActivity : AppCompatActivity(), Presenter {
         addLongPressOrderableListComponent()
         addHandleOrderableListComponent()
         addReorderComponentGroupComponent()
+        addReorderAllButLast()
     }
 
     private fun addUnorderableListComponent() {
@@ -76,7 +78,14 @@ class ReorderListActivity : AppCompatActivity(), Presenter {
 
     private fun addReorderComponentGroupComponent() {
         componentController.addComponent(LabeledComponent("Reorder in ComponentGroup"))
-        componentController.addComponent(ReorderableComponentGroup())
+        componentController.addComponent(ReorderableComponentGroup().apply {
+            setEndGap(20)
+        })
+    }
+
+    private fun addReorderAllButLast() {
+        componentController.addComponent(LabeledComponent("Reorder all but the last, fixed, item"))
+        componentController.addComponent(ReorderAllButLastComponent())
     }
 
     override fun onStartDrag(viewHolder: ComponentViewHolder<Presenter, String>) {
@@ -115,5 +124,28 @@ class ReorderableComponentGroup : ComponentGroup() {
         }
     }
 
-    override fun isReorderable() = true
+    override fun canPickUpItem(index: Int) = true
+}
+
+class ReorderAllButLastComponent : Component() {
+
+    override fun getPresenter(position: Int) = Unit
+
+    override fun getItem(position: Int) = if (position == count - 1) {
+        "Last Item"
+    } else {
+        position.toString()
+    }
+
+    override fun getCount() = 10
+
+    override fun getHolderType(position: Int) = LabeledComponentViewHolder::class.java
+
+    override fun canDropItem(fromComponent: Component?, fromIndex: Int, toIndex: Int): Boolean {
+        return fromComponent == this && toIndex != count - 1
+    }
+
+    override fun canPickUpItem(index: Int): Boolean {
+        return index != count - 1
+    }
 }
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
@@ -111,17 +111,23 @@ class ReorderViewHolder : ComponentViewHolder<Presenter, String>() {
 
 class ReorderAllButLastComponent : Component() {
 
+    val data = (0 until 10).toMutableList()
+
     override fun getPresenter(position: Int) = Unit
 
     override fun getItem(position: Int) = if (position == count - 1) {
         "Last Item. Can't move"
     } else {
-        position.toString()
+        data[position].toString()
     }
 
-    override fun getCount(): Int = 10
+    override fun getCount(): Int = data.size
 
     override fun getHolderType(position: Int) = LabeledComponentViewHolder::class.java
+
+    override fun onItemsMoved(fromIndex: Int, toIndex: Int) {
+        data.add(toIndex, data.removeAt(fromIndex))
+    }
 
     override fun canDropItem(fromIndex: Int, toIndex: Int): Boolean {
         return toIndex != count - 1
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
@@ -2,7 +2,6 @@ package com.yelp.android.bentosampleapp
 
 import android.os.Bundle
 import android.util.Log
-import android.view.LayoutInflater
 import android.view.MotionEvent
 import android.view.View
 import android.view.ViewGroup
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
@@ -32,7 +32,6 @@ class ReorderListActivity : AppCompatActivity(), Presenter {
         addUnorderableListComponent()
         addLongPressOrderableListComponent()
         addHandleOrderableListComponent()
-        addReorderComponentGroupComponent()
         addReorderAllButLast()
     }
 
@@ -76,18 +75,6 @@ class ReorderListActivity : AppCompatActivity(), Presenter {
         componentController.addComponent(componentGroup)
     }
 
-    private fun addReorderComponentGroupComponent() {
-        componentController.addComponent(LabeledComponent("Reorder in ComponentGroup"))
-        componentController.addComponent(
-                ComponentGroup().apply {
-                    addComponent(ComponentGroup().apply {
-                        addComponent(ReorderableComponentGroup().apply {
-                            setEndGap(20)
-                        })
-                    })
-                })
-    }
-
     private fun addReorderAllButLast() {
         componentController.addComponent(LabeledComponent("Reorder all but the last, fixed, item"))
         componentController.addComponent(ReorderAllButLastComponent())
@@ -122,28 +109,22 @@ class ReorderViewHolder : ComponentViewHolder<Presenter, String>() {
     }
 }
 
-class ReorderableComponentGroup : ComponentGroup() {
-    init {
-        for (number in (0..10)) {
-            addComponent(LabeledComponent("Component #$number"))
-        }
+class ReorderAllButLastComponent : Component() {
+
+    override fun getPresenter(position: Int) = Unit
+
+    override fun getItem(position: Int) = if (position == count - 1) {
+        "Last Item. Can't move"
+    } else {
+        position.toString()
     }
 
-    override fun canPickUpItem(index: Int) = true
-}
+    override fun getCount(): Int = 10
 
-class ReorderAllButLastComponent : ComponentGroup() {
+    override fun getHolderType(position: Int) = LabeledComponentViewHolder::class.java
 
-    init {
-        for (i in (0 until 10)) {
-            addComponent(LabeledComponent(i.toString()))
-        }
-
-        addComponent(LabeledComponent("Last item"))
-    }
-
-    override fun canDropItem(fromComponent: Component?, fromIndex: Int, toIndex: Int): Boolean {
-        return fromComponent == this && toIndex != count - 1
+    override fun canDropItem(fromIndex: Int, toIndex: Int): Boolean {
+        return toIndex != count - 1
     }
 
     override fun canPickUpItem(index: Int): Boolean {
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
@@ -11,15 +11,13 @@ import androidx.appcompat.app.AppCompatActivity
 import com.yelp.android.bento.componentcontrollers.RecyclerViewComponentController
 import com.yelp.android.bento.components.ListComponent
 import com.yelp.android.bento.components.OnItemMovedCallback
+import com.yelp.android.bento.core.ComponentGroup
 import com.yelp.android.bento.core.ComponentViewHolder
 import com.yelp.android.bentosampleapp.components.LabeledComponent
 import com.yelp.android.bentosampleapp.components.LabeledComponentViewHolder
 import kotlinx.android.synthetic.main.activity_recycler_view.*
 
 class ReorderListActivity : AppCompatActivity(), Presenter {
-
-    private lateinit var handleComponent: ListComponent<Presenter, ReorderElement>
-    private lateinit var longPressComponent: ListComponent<Unit, String>
 
     private val componentController by lazy {
         RecyclerViewComponentController(recyclerView)
@@ -45,7 +43,7 @@ class ReorderListActivity : AppCompatActivity(), Presenter {
 
     private fun addLongPressOrderableListComponent() {
         componentController.addComponent(LabeledComponent("Long press to reorder"))
-        longPressComponent =
+        val longPressComponent =
                 ListComponent<Unit, String>(Unit, LabeledComponentViewHolder::class.java, 2)
         longPressComponent.setIsReorderable(true)
         longPressComponent.toggleDivider(false)
@@ -59,23 +57,25 @@ class ReorderListActivity : AppCompatActivity(), Presenter {
     }
 
     private fun addHandleOrderableListComponent() {
-        componentController.addComponent(LabeledComponent("Drag handle to reorder"))
-        handleComponent =
-                ListComponent<Presenter, ReorderElement>(this, ReorderViewHolder::class.java, 2)
+        val handleComponent =
+                ListComponent<Presenter, String>(this, ReorderViewHolder::class.java, 2)
         handleComponent.setIsReorderable(true)
         handleComponent.toggleDivider(false)
-        handleComponent.setData((0..10).map { ReorderElement('A'.plus(it).toString(), it) })
-        handleComponent.setOnItemMovedCallback(object : OnItemMovedCallback<ReorderElement> {
-            override fun onItemMoved(oldIndex: Int, newIndex: Int, newData: List<ReorderElement>) {
-                newData.forEachIndexed { index, reorderElement -> reorderElement.index = index }
+        handleComponent.setData((0..10).map { 'A'.plus(it).toString() })
+        handleComponent.setOnItemMovedCallback(object : OnItemMovedCallback<String> {
+            override fun onItemMoved(oldIndex: Int, newIndex: Int, newData: List<String>) {
                 Log.i("Reordered", "New list ordering: $newData")
             }
         })
-        componentController.addComponent(handleComponent)
+        val componentGroup = ComponentGroup().apply {
+            addComponent(LabeledComponent("Drag handle to reorder"))
+            addComponent(handleComponent)
+        }
+        componentController.addComponent(componentGroup)
     }
 
     override fun onStartDrag(position: Int) {
-        componentController.onItemPickedUp(handleComponent, position)
+        componentController.onItemPickedUp(position)
     }
 }
 
@@ -83,11 +83,10 @@ interface Presenter {
     fun onStartDrag(position: Int)
 }
 
-class ReorderViewHolder : ComponentViewHolder<Presenter, ReorderElement>() {
+class ReorderViewHolder : ComponentViewHolder<Presenter, String>() {
 
     private lateinit var text: TextView
     private lateinit var presenter: Presenter
-    private lateinit var element: ReorderElement
 
     override fun inflate(parent: ViewGroup): View {
         return LayoutInflater.from(parent.context)
@@ -95,25 +94,15 @@ class ReorderViewHolder : ComponentViewHolder<Presenter, ReorderElement>() {
                     text = findViewById(R.id.textview)
                     findViewById<View>(R.id.handle).setOnTouchListener { _, event ->
                         if (event.action == MotionEvent.ACTION_DOWN) {
-                            presenter.onStartDrag(element.index)
+                            presenter.onStartDrag(absolutePosition)
                         }
                         false
                     }
                 }
     }
 
-    override fun bind(presenter: Presenter, element: ReorderElement) {
-        text.text = element.text
-        this.element = element
+    override fun bind(presenter: Presenter, element: String) {
+        text.text = element
         this.presenter = presenter
     }
-}
-
-data class ReorderElement(
-        val text: String,
-        var index: Int
-)
-
-fun <T> MutableList<T>.move(oldIndex: Int, newIndex: Int) {
-    add(newIndex, removeAt(oldIndex))
 }
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
@@ -78,9 +78,14 @@ class ReorderListActivity : AppCompatActivity(), Presenter {
 
     private fun addReorderComponentGroupComponent() {
         componentController.addComponent(LabeledComponent("Reorder in ComponentGroup"))
-        componentController.addComponent(ReorderableComponentGroup().apply {
-            setEndGap(20)
-        })
+        componentController.addComponent(
+                ComponentGroup().apply {
+                    addComponent(ComponentGroup().apply {
+                        addComponent(ReorderableComponentGroup().apply {
+                            setEndGap(20)
+                        })
+                    })
+                })
     }
 
     private fun addReorderAllButLast() {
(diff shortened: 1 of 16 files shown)


bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
```diff
@@ -253,6 +255,21 @@ public int getSpanSize(int position) {
...
     }
 
+    @Override
+    public void onItemsMoved(int oldIndex, int newIndex) {
+        super.onItemsMoved(oldIndex, newIndex);
+
+        mData.add(newIndex, mData.remove(oldIndex));
+
+        if (mOnItemMovedCallback != null) {
+            mOnItemMovedCallback.onItemMoved(oldIndex, newIndex, mData);
```
wyattjsmith1: @dwaxemberg Is it ok to be passing back a raw `mData`, or should I be passing a copy of the list?

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+        componentController.addComponent(list)
+    }
+
+    private fun addLongPressOrderableListComponent() {
+        componentController.addComponent(LabeledComponent("Long press to reorder"))
+        val longPressComponent =
+                ListComponent<Unit, String>(Unit, LabeledComponentViewHolder::class.java, 2)
+        longPressComponent.setIsReorderable(true)
+        longPressComponent.toggleDivider(false)
+        longPressComponent.setData((0..10).map { 'a'.plus(it).toString() })
```
dwaxemberg, line 51: This would be cleaner with string interpolation: "a $it". Also, why "a"? Can it be something like "Item $it"?

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+    fun onStartDrag(position: Int)
+}
+
+class ReorderViewHolder : ComponentViewHolder<Presenter, String>() {
+
+    private lateinit var text: TextView
+    private lateinit var presenter: Presenter
+
+    override fun inflate(parent: ViewGroup): View {
+        return LayoutInflater.from(parent.context)
```
dwaxemberg: Can be replaced with `parent.inflate()`

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
dwaxemberg: Generally it's bad practice unless the list is immutable. You could use Collections.immutableList or convert this to Kotlin and use the immutable List. 

Aside: Why are we passing the list back at all?

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
```diff
@@ -253,12 +255,42 @@ public int getSpanSize(int position) {
...
+        }
+    }
+
+    public void setOnItemMovedCallback(OnItemMovedCallback<T> callback) {
```
dwaxemberg, line 271: Move the public methods above the private/protected ones

bento/src/main/java/com/yelp/android/bento/core/Component.java
```diff
@@ -262,6 +256,20 @@ final Object getPresenterInternal(int position) {
...
+     * during {@link OnTouchListener#onTouch(android.view.View, android.view.MotionEvent)} if
+     * {@link android.view.MotionEvent} is {@link MotionEvent#ACTION_DOWN}.
+     *
+     * @see Component#onItemsMoved(int, int)
+     */
+    public boolean isReorderable() {
+        return false;
+    }
+
+
```
dwaxemberg: remove extra empty line

bento/src/main/java/com/yelp/android/bento/core/Component.java
```diff
@@ -298,6 +306,32 @@ public void setSpanSizeLookup(@NonNull SpanSizeLookup lookup) {
...
 
+    /**
+     * Override this method to handle reordering of items.
+     *
+     * @param oldIndex The index the item was originally in.
+     * @param newIndex The index the item was moved to.
+     * @see Component#isReorderable() (boolean)
+     */
+    public void onItemsMoved(int oldIndex, int newIndex) {
+
```
dwaxemberg: Remove empty line

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -30,8 +35,11 @@
 import kotlin.sequences.Sequence;
 import org.jetbrains.annotations.Nullable;
 
-/** Implementation of {@link ComponentController} for {@link RecyclerView}s. */
-public class RecyclerViewComponentController implements ComponentController {
+/**
+ * Implementation of {@link ComponentController} for {@link RecyclerView}s.
+ */
+public class RecyclerViewComponentController implements ComponentController,
+        OnItemMovedPositionListener {
```
dwaxemberg, line 42: What about other ComponentControllers?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -268,6 +283,27 @@ public void scrollToComponentWithOffset(@NonNull Component component, int offset
         }
     }
 
+    @Override
+    public void onItemMovedPosition(int oldIndex, int newIndex) {
+        RangedValue<Component> rangedValue = mComponentGroup.getLowestRangeValue(oldIndex);
+        rangedValue.mValue.onItemsMoved(oldIndex - rangedValue.mRange.mLower,
+                newIndex - rangedValue.mRange.mLower);
+
+        // bind is not called again, so we need to go through and properly set all the positions
```
dwaxemberg: Capital first letter and period at the end.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -268,6 +283,27 @@ public void scrollToComponentWithOffset(@NonNull Component component, int offset
...
+        int currentIndex = Math.min(oldIndex, newIndex);
+        int highIndex = Math.max(oldIndex, newIndex);
+        while (currentIndex <= highIndex) {
+            ((ViewHolderWrapper) mRecyclerView.findViewHolderForAdapterPosition(currentIndex)).mViewHolder.setAbsolutePosition(currentIndex++);
```
dwaxemberg: This line is way too long.

Also, we should just iterate through the visible view holders, all others will have bind called on them when the view scrolls.

bento/src/main/java/com/yelp/android/bento/components/OnItemMovedCallback.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+interface OnItemMovedCallback<T> {
+
+    /**
+     * Called when an item has been dropped somewhere else in the list.
+     * @param oldIndex The old index of the item.
+     * @param newIndex The index where the item landed.
+     * @param newData A new list of data after the reorder.
+     */
+    fun onItemMoved(oldIndex: Int, newIndex: Int, newData: List<T>)
```
dwaxemberg: This needs to be used by all components, not just list components therefore we shouldn't make `newData` a list. instead ListComponent should use a `OnItemMovedCallback<List<T>>` in which case `newData` becomes `List<T>`.

But as I mentioned somewhere else, why are we passing the data through this listener?

bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    override fun onSwiped(viewHolder: RecyclerView.ViewHolder, direction: Int) {
+        // No swipe support for now.
+    }
+
+    override fun clearView(recyclerView: RecyclerView, viewHolder: RecyclerView.ViewHolder) {
+        super.clearView(recyclerView, viewHolder)
+
+        // This simply checks when the user "drops" an item. onMove is called anytime the item is
```
dwaxemberg, line 71: This says "onMove is called anytime the item..." but it's in the `clearView` method. Can you elaborate a bit?

bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    private var dragFrom = -1
+    private var dragTo = -1
+
+    override fun canDropOver(
+            recyclerView: RecyclerView,
+            current: RecyclerView.ViewHolder,
+            target: RecyclerView.ViewHolder
+    ): Boolean {
+        // Only allow reorder if it is within the same component.
```
dwaxemberg, line 22: What if I have a ComponentGroup full of Components with 1 view and I want to swap them?
For example, if I wanted to move items inside a CarouselComponent or some sort of grid-like ComponentGroup.

bento/src/main/java/com/yelp/android/bento/core/ComponentViewHolder.kt
```diff
@@ -16,6 +16,12 @@ import android.view.ViewGroup
  */
 abstract class ComponentViewHolder<P, T> {
 
+    /**
+     * Contains the absolute position within the entire [ComponentController]. This should only be
+     * set by the [ComponentController].
+     */
+    var absolutePosition: Int = -1
```
dwaxemberg, line 24: Give this an internal setter and public getter to guarantee only the ComponentController (or at least only the library) can set it.

bento/src/main/java/com/yelp/android/bento/core/Component.java
```diff
@@ -184,17 +188,6 @@ public int getNumberLanes() {
...
     }
 
-    /**
-     * Returns the number of lanes at a given position. This will usually return the same as
-     * {@link Component#getNumberLanes()}, but {@link ComponentGroup}s will want to return the
-     * number of lanes at a given index. For example, if you have a {@link ComponentGroup} with
-     * two {@link Component}s with a different number of lanes, this should return the number of
-     * lanes in the component that owns the position.
-     */
-    public int getNumberLanesAtPosition(int position) {
```
dwaxemberg, line 194: Why are we removing this? It seems unrelated to the changes being made.

bento/src/main/java/com/yelp/android/bento/core/Component.java
```diff
@@ -298,6 +306,32 @@ public void setSpanSizeLookup(@NonNull SpanSizeLookup lookup) {
...
+    }
+
+    /**
+     * Returns the lowest component in the component tree. In general, a {@link ComponentGroup} will
+     * recursively call this method, removing the lower bound of its range each time. A
+     * {@link Component} should just return itself.
+     * @param index The index to search for.
+     * @return The lowest component (leaf) in the component tree.
+     */
+    public Component getLowestComponentAtIndex(int index) {
```
dwaxemberg: Instead of putting this in the Component class, I think we should do something similar to the `findComponentOffset` in ComponentGroup where we just recursively search the children.

dwaxemberg: Also, what if `index` is not `0 <= index < count`

bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java
```diff
@@ -412,6 +407,41 @@ public int findComponentOffset(@NonNull Component component) {
...
+        Range range = rangeOf(componentAtPos);
+        if (range == null) {
+            throw new IllegalStateException("Found component, but no range associated.");
+        }
+
+        return componentAtPos.getLowestComponentAtIndex(index - range.mLower);
+    }
+
+    /**
+     * Similar to {@link #getLowestComponentAtIndex(int)}, but returns both the component and the
```
dwaxemberg: If these methods do the same thing, couldn't we have `getLowestComponentAtIndex(int)` return `getLowestRangeValue(int index).mValue`?

bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java
```diff
@@ -412,6 +407,41 @@ public int findComponentOffset(@NonNull Component component) {
...
+     * absolute range within the controller.
+     *
+     * @param index The index to search for.
+     * @return Both a component and an absolute range over the entire controller.
+     */
+    @Override
+    public RangedValue<Component> getLowestRangeValue(int index) {
```
dwaxemberg: This method seems very similar to `findComponentOffset` except it includes the upper bound as well.

dwaxemberg (changes requested):

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
wyattjsmith1: So, `Char.plus()` doesn't concatenate, it adds. Basically, I want to distinguish between each section, so one section is numbers, one is lowercase, and one is uppercase. The `plus` function adds to the Char's ascii value. It is the Java equivalent to `'a' + 1 = 'b'`.

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
wyattjsmith1: Ah, this was needed at one time, but not since we added the absolute position to the ViewHolders. I'll remove it.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
wyattjsmith1: I was thinking about this. In my opinion, there isn't really a good way to implement this in other controllers. How would a `ViewPager` reorder? Android doesn't really have a built in way to do that.

Additionally, the state of being reorderable is managed by the Component, not the Controller. In that way, Components are really requesting to be reorderable, but it is up to the Controller to reorder. Not sure if this is what we are looking for.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
wyattjsmith1: Hmm... ok. How would we iterate only the visible view holders?

bento/src/main/java/com/yelp/android/bento/components/OnItemMovedCallback.kt
wyattjsmith1: Removed in comment above.

bento/src/main/java/com/yelp/android/bento/core/Component.java
wyattjsmith1: Eh, a little unrelated, but still related. I find that both grids and reordering require me to be able to find a component at a given absolute index. I just created a getLowestComponentAtIndex, then call whatever method from there. Calling `getLowestComponentAtIndex(n).getNumberLanes()` seems cleaner than having a bunch of redundant methods that recursively find the lowest component, then return some field from it.

bento/src/main/java/com/yelp/android/bento/core/ComponentViewHolder.kt
wyattjsmith1: The problem is that doesn't compile. I think it is because the RecyclerViewComponentController is in java, which doesn't recognize internal, and it is probably package protected then.

bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java
wyattjsmith1: You aren't wrong. `findComponentOffset` takes a component and  `getLowestRangeValue` takes an index. Are you suggesting something?

bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
wyattjsmith1: I agree that this implementation is limiting. I guess my questions is, do you have any other suggestions? I suppose we could have some default implementation of this in `Component` that does this check, and override it in `ComponentGroup` if needed. Do you think that would be better?

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt
```diff
@@ -0,0 +1,107 @@
...
+
+class ReorderViewHolder : ComponentViewHolder<Presenter, String>() {
+
+    private lateinit var text: TextView
+    private lateinit var presenter: Presenter
+
+    override fun inflate(parent: ViewGroup): View {
+        return parent.inflate<View>(R.layout.reorderable_view_holder).also {
+                    text = it.findViewById(R.id.textview)
+                    it.findViewById<View>(R.id.handle).setOnTouchListener { _, event ->
```
wyattjsmith1: @dwaxemberg Do you think it would be useful to add some sort of extension function such as `setOnDragStartListener()` that would take a lambda? It feels weird to add a touch listener and listen for `ACTION_DOWN`. It could be used like so:

```
view.setOnDragStartListener {
  presenter.onStartDrag(this)
}

File: bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt (hunk 12 above)
Comment (wyattjsmith1) [reply bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt:22]: Another thing to remember is that the `RecyclerView`'s reordering only works with `ViewHolder`s. If we were to allow entire components to be reordered, we would have to check that their count is only 1 or some weird things might happen.

Comment (wyattjsmith1) [reply bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt:22]: Another thing I'm gonna throw out is that we may not want to allow everything within one component be reorderable. Consider this layout. It is a ListComponent, where we add another item to the end to view/add photos. We probably don't want to be able to rearrange a photo behind the button.
image

File: bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java (hunk 9 above)
Comment (dwaxemberg) [reply bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java]: You can ask the layout manager for the first and last visible indices. We do a similar thing in the VisibilityListener.

File: bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt (hunk 12 above)
Comment (dwaxemberg) [reply bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt:22]: Yeah I can see how this would get complicated, maybe it should be a follow up. I do think we should support reordering of components within a component group though. We would definitely need to figure out a way to guarantee these are single item components to avoid any messy RecyclerView issues. 

I guess for now just forget about it and maybe open an issue as "nice to have".

File: bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java (hunk 7 above)
Comment (dwaxemberg) [reply bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java:42]: I was thinking of a ListView primarily, but that might also be difficult to handle. Maybe we should just stick to RecylerViews for now.

File: bento/src/main/java/com/yelp/android/bento/core/Component.java (hunk 14 above)
Comment (dwaxemberg) [reply bento/src/main/java/com/yelp/android/bento/core/Component.java:194]: Gotcha. I don't like the term "lowest" to refer to the leaf node however. Also, since this is not just returning a field, I think we should change to "find" instead of "get" so it doesn't give the impression that this is a costless operation. What do you think about renaming it to something like `findIndexOwner` or `findComponentWithIndex`?

File: bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java (hunk 17 above)
Comment (dwaxemberg) [reply bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java]: Could we combine the two methods somehow? (e.g. make a helper method that both methods call into in order to reduce some of the very similar logic)

Comment (dwaxemberg) [reply bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java]: Or maybe just give them complimentary names (like `findComponentWithIndex` and `findIndexOfComponent` or something)

File: bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt (hunk 18 above)
Comment (dwaxemberg) [reply bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt]: Generally I like the idea, but would it be useful to have it in the library vs. making the consumers declare that function if they want it?

Comment (wyattjsmith1) [reply bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ReorderListActivity.kt]: Spoke on Slack, we agreed it makes sense in the Bento library.

File: bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt (hunk 12 above)
Comment (wyattjsmith1) [reply bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt:22]: Whoops, looks like Github lost my comment. TBH, I  think this would be difficult to patch in later. Additionally, the current implementation would not work with the example above, which is the primary reason I am implementing this.

One possible solution is to create a function `canDropItem(fromComponent, fromIndex, toComponent, toIndex)` in each `Component` that allows us to control whether a `Component` can take a given item. 

This solution would cover the case where a component doesn't want to allow an item to be dropped in a certain location (like my example above), but does not consider a `ComponentGroup` with `Component`s with a count > 1. It seems like the only way a `ComponentGroup` could be reorderable is if all of its children are `Component`s with a count == 1. I suppose we could do this check, but it wouldn't be the greatest. Additionally, what would we do when a child has a count > 1? Throw an exception? Not reorder? Just roll with it and hope for the best?

Comment (wyattjsmith1) [reply bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt:22]: Ok, I added a solution for this. The only downside I see if that we now have `findRangedComponentWithIndex` and `findReorderTargetAtIndex` which have very similar implementations, but I don't see a way to combine them.

File: bento/src/main/java/com/yelp/android/bento/core/Component.java [hunk 19]
```diff
@@ -298,6 +303,27 @@ public void setSpanSizeLookup(@NonNull SpanSizeLookup lookup) {
...
+     *
+     * @param oldIndex The index the item was originally in.
+     * @param newIndex The index the item was moved to.
+     * @see Component#isReorderable() (boolean)
+     */
+    public void onItemsMoved(int oldIndex, int newIndex) {
+    }
+
+    /**
+     * Checks if a n item from one component can be dropped in this component at a given index.
```
dwaxemberg: typo: remove the space before the 'n' in 'an'

bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java
```diff
@@ -412,6 +407,58 @@ public int findComponentOffset(@NonNull Component component) {
...
+     * Finds and returns the component at lowest level (leaf) that encompasses the index.
+     *
+     * @param index The index to search for.
+     * @return The lowest component in the tree.
+     */
+    public Component findComponentWithIndex(int index) {
+        return findRangedComponentWithIndex(index).mValue;
+    }
+
+    public RangedValue<Component> findReorderTargetAtIndex(int index) {
```
dwaxemberg: Add a javadoc for this

bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    override fun getMovementFlags(
+            recyclerView: RecyclerView,
+            viewHolder: RecyclerView.ViewHolder
+    ): Int {
+        val lowestComponent = component.findReorderTargetAtIndex(viewHolder.adapterPosition).mValue
+        if (lowestComponent.isReorderable) {
+            if (lowestComponent is ComponentGroup) {
+                for (index in (0 until lowestComponent.size)) {
+                    if (lowestComponent[index].count != 1) {
```
dwaxemberg: Can we use `asSequence` here?
```
if (lowestComponent.asSequence().any { it.count != 1 }) {
...
```

Also, I think we should check >0 instead of != 1 because if we have a bunch of components with a count of 0 we should be able to safely ignore them.

wyattjsmith1: I assume you mean > 1?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/ListViewComponentController.kt
```diff
@@ -15,6 +15,7 @@ import com.yelp.android.bento.core.ComponentController
 import com.yelp.android.bento.core.ComponentGroup
 import com.yelp.android.bento.core.ComponentViewHolder
 import com.yelp.android.bento.core.ComponentVisibilityListener
+import com.yelp.android.bento.core.GapViewHolder
```
wyattjsmith1: Whoops, removing.

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
```diff
@@ -253,6 +255,38 @@ public int getSpanSize(int position) {
         };
     }
 
+    @Override
+    public void onItemsMoved(
```
dwaxemberg: Should we make this `final` here?

bento/src/main/java/com/yelp/android/bento/components/OnItemMovedCallback.kt
```diff
@@ -0,0 +1,14 @@
+package com.yelp.android.bento.components
+
+/**
+ * Interface for listening to drag and drop events.
+ */
+interface OnItemMovedCallback<T> {
+
+    /**
+     * Called when an item has been dropped somewhere else in the list.
```
dwaxemberg: nit: change "list" to "component"

bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java
```diff
@@ -40,6 +44,9 @@ public ComponentGroup() {
         mSpanSizeLookup = new SpanSizeLookup() {
             @Override
             public int getSpanSize(int position) {
+                if (hasGap(position)) {
```
dwaxemberg: This looks like a bug fix for something. Is it related to this PR or should we make it separate?

bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java
```diff
@@ -412,6 +410,43 @@ public int findComponentOffset(@NonNull Component component) {
...
+    }
+
+    /**
+     * Returns both the component and the absolute range within the controller.
+     *
+     * @param index The index to search for.
+     * @return Both a component and an absolute range over the entire controller.
+     */
+    public RangedValue<Component> findRangedComponentWithIndex(int index) {
+
```
dwaxemberg: remove extra line

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
wyattjsmith1: Eh, I can't imagine a reason someone would want to override this.

bento/src/main/java/com/yelp/android/bento/core/ComponentGroup.java
wyattjsmith1: Eh, back when we were testing `ComponentGroups` with gaps, it crashed and this fixed it. Now that we eliminated that feature, the test app works. I will open a separate PR for this.

dwaxemberg (approved):

dwaxemberg:
Don't forget to bump the version

PR #20, opened 2019-07-11
Added check for gap in ComponentGroup

wyattjsmith1:
There is a bug where adding a gap to headers can cause an exception to be thrown when calculating a component span. This commit checks to see if the requested index is a gap, then sets the span to the component's number of columns.


dwaxemberg (approved):

dwaxemberg:
Don't forget to bump the version

PR #21, opened 2019-07-16
More fine-grained control over data in ListComponent

zachbryant:
This commit improves the flexibility of ListComponent by enabling more fine-grained updates to the data. The return type of ListComponent.removeData(T data) now returns the int value of the index removed. Additionally a new method to insert data at a certain index in the existing data has been added. The user should be able to replace or re-insert data into the list without re-rendering the entire list to change one item.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .*
 *.iml
+.idea
 
 gradle.log
 local.properties
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ListViewActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ListViewActivity.kt
@@ -1,6 +1,7 @@
 package com.yelp.android.bentosampleapp
 
 import android.os.Bundle
+import android.util.Log
 import android.view.Menu
 import android.view.MenuItem
 import android.widget.ArrayAdapter
@@ -44,6 +45,7 @@ class ListViewActivity : AppCompatActivity() {
         addCarouselComponent(controller)
         addArrayAdapterComponent(controller)
         addAnimatedComponent(controller)
+        addListComponentRemovableItems(controller)
     }
 
     override fun onCreateOptionsMenu(menu: Menu): Boolean {
@@ -79,14 +81,33 @@ class ListViewActivity : AppCompatActivity() {
         controller.addComponent(simpleComponent)
     }
 
-
-    private fun addListComponent(controller: ComponentController) {
+    private fun addListComponent(controller: ComponentController, removable: Boolean = false) {
         with(ListComponent(null,
                 ListComponentExampleViewHolder::class.java)) {
             setStartGap(50)
             setData((1 until 42).map { "List element $it" })
             toggleDivider(false)
             controller.addComponent(this)
+        }
+    }
+
+    private fun addListComponentRemovableItems(controller: ComponentController) {
+        val component = ListComponent(null, ListComponentExampleViewHolder::class.java)
+        with(component) {
+            setStartGap(50)
+            setData(listOf("List element 1"))
+            toggleDivider(true)
+            controller.addComponent(this)
+
+            listOf(1, 1, 1, 2, 4).forEachIndexed { index, i -> insertData(i, "List element ${index + 2}") }
+            insertData(0, "List element First")
+            insertData(0, "List element First")
+            insertData(1 + component.count / 2, "List element Last")
+            removeData("List element First")
+            val index = removeData("List element 1")
+            insertData(index, "List element 1")
+            removeData("List element 2")
+            removeData("List element nonexistent")
         }
     }
 
--- a/bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
+++ b/bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
@@ -89,14 +89,16 @@ public class ListComponent<P, T> extends Component {
      * Removes the provided data items from the list.
      *
      * @param data The data item to remove from the list.
+     * @return an int representing the index that the removed data occupied
      */
-    public void removeData(@NonNull T data) {
+    public int removeData(@NonNull T data) {
         int index = mData.indexOf(data);
         // Check if the object indeed is in the list.
         if (index != -1) {
             mData.remove(index);
             notifyItemRangeRemoved(getRemoveIndexStart(index), getRemoveItemCount());
         }
+        return index;
     }
 
     /**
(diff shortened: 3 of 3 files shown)


redwarp:
@zachbryant that would be a great addition! It seems though that this change does not take into account the presence or not of dividers. If the ListComponent has separator, then it also displays some grey line between each element. In effet, it displays some extra items, which will impact the index of the item being added. Check https://github.com/Yelp/bento/blob/master/bento/src/main/java/com/yelp/android/bento/components/ListComponent.java#L105

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
```diff
@@ -85,18 +85,31 @@ public void appendData(@NonNull List<T> data) {
...
 
+    /**
+     * Inserts the specified data at the specified index in the component data.
+     *
+     * @param index the index to insert the given data
+     * @param data the data to be inserted
+     */
+    public void insertData(int index, @NonNull T data) {
+        mData.add(index, data);
+        notifyItemRangeInserted(index, index + 1);
```
dwaxemberg: This needs to take dividers into account. Pretty sure the first argument is fine, but the second needs to be either +1 or +2 if there is a divider.

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
```diff
@@ -85,18 +85,31 @@ public void appendData(@NonNull List<T> data) {
...
+    }
+
     /**
      * Removes the provided data items from the list.
      *
      * @param data The data item to remove from the list.
+     * @return an int representing the index that the removed data occupied
```
dwaxemberg: nit: remove "an int representing". It's redundant due to the function signature.
Also, add something about returning -1 when nothing is removed.

dwaxemberg (changes requested):

dwaxemberg:
bump

zachbryant:
Thanks for the reminder! Things are currently crazy at school but I'll have time to revise my request after Halloween

redwarp:
@zachbryant bump?

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
zachbryant: Got it! Looks like as long as the itemCount is right, the index doesn't seem to matter. Might look into that another time

zachbryant:
> @zachbryant that would be a great addition! It seems though that this change does not take into account the presence or not of dividers. If the ListComponent has separator, then it also displays some grey line between each element. In effet, it displays some extra items, which will impact the index of the item being added. Check https://github.com/Yelp/bento/blob/master/bento/src/main/java/com/yelp/android/bento/components/ListComponent.java#L105

@redwarp I noticed that no matter what index is given, as long as the item count is correct the whole list re-renders. Not sure if that's intended.. but passing just `index` seems to be in line with `appendData` too, unless I'm missing something? Let me know and I'll change it to an adjusted index

.gitignore
```diff
@@ -1,5 +1,6 @@
 .*
 *.iml
+.idea
```
redwarp, line 3: I don't think this should be there

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ListViewActivity.kt
```diff
@@ -1,6 +1,7 @@
 package com.yelp.android.bentosampleapp
 
 import android.os.Bundle
+import android.util.Log
```
redwarp, line 4: This should not be there, Log is not used

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ListViewActivity.kt
```diff
@@ -79,8 +81,7 @@ class ListViewActivity : AppCompatActivity() {
         controller.addComponent(simpleComponent)
     }
 
-
-    private fun addListComponent(controller: ComponentController) {
+    private fun addListComponent(controller: ComponentController, removable: Boolean = false) {
```
redwarp, line 84: the removable flag is not used

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ListViewActivity.kt
```diff
@@ -90,6 +91,26 @@ class ListViewActivity : AppCompatActivity() {
...
+
+            listOf(1, 1, 1, 2, 4).forEachIndexed { index, i -> insertData(i, "List element ${index + 2}") }
+            insertData(0, "List element First")
+            insertData(0, "List element First")
+            insertData(1 + component.count / 2, "List element Last")
+            removeData("List element First")
+            val index = removeData("List element 1")
+            insertData(index, "List element 1")
+            removeData("List element 2")
+            removeData("List element nonexistent")
```
redwarp, line 110: I don't think this is a really useful code: you are adding and removing items even before the component would be displayed on screen. You might want to consider instead first displaying a list of items, and adding a button somewhere to try the scenario where you remove items dynamically.

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/ListViewActivity.kt
```diff
@@ -1,6 +1,7 @@
 package com.yelp.android.bentosampleapp
```
redwarp, line 1: Be aware that the ListViewActivity was created to verify that list views display stuff properly. There is no link between the ListViewActivity and the ListComponent. Bento is compatible with both ListViews and RecyclerViews, ViewPagers...

bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
```diff
@@ -86,17 +86,31 @@ public void appendData(@NonNull List<T> data) {
...
-     * Removes the provided data items from the list.
+     * Inserts a single list item at the specified index in the list.
+     *
+     * @param index The index used to insert the given data in the list.
+     * @param data The list item to be inserted.
+     */
+    public void insertData(int index, @NonNull T data) {
+        mData.add(index, data);
+        // Update 2 items if dividers are showing.
+        notifyItemRangeInserted(index, mShouldShowDivider ? 2 : 1);
```
redwarp, line 97: if divider are visible, you probably need to multiply index by 2, otherwise you will have some surprises

redwarp (changes requested):
Please add tests to ListComponentTest to verify that proper index are updated with and without dividers.

redwarp:
@zachbryant is this still alive?

PR #27, opened 2019-07-31
Specify the bento_ prefix in the gradle configuration

redwarp:
Future resources will have to begin with bento_ as well.


redwarp:
Rebased on master

PR #29, opened 2019-08-13
Save scroll position of the CarouselComponent when scrolled offscreen

argo49:
- Added two new fields to the `CarouselViewModel` to save scroll position
- Made the viewModel a member of the CarouselComponent so it can remember where it was
- Updated the contributing steps to include updating the CHANGELOG.md file because I saw it was a missing step
- Added tests for saving and restoring scroll position to the carousel

### DIFF
--- a/bento/src/main/java/com/yelp/android/bento/components/CarouselComponent.kt
+++ b/bento/src/main/java/com/yelp/android/bento/components/CarouselComponent.kt
@@ -75,20 +75,6 @@ open class CarouselComponentViewHolder : ComponentViewHolder<Unit?, CarouselView
     lateinit var element: CarouselViewModel
     private var attachedPool = false
 
-    /**
-     * Saves the scroll position of the carousel to the view model so that it can be restored when
-     * the carousel recyclerView itself is recycled in the larger component list. Scroll position
-     * is saved when the scroll state of the carousel transitions to
-     * [RecyclerView.SCROLL_STATE_IDLE].
-     */
-    private fun saveScrollPosition() {
-        (recyclerView.layoutManager as? LinearLayoutManager)?.apply {
-            element.scrollPosition = findFirstVisibleItemPosition()
-            element.scrollPositionOffset =
-                    recyclerView.getChildAt(0).left - recyclerView.paddingLeft
-        }
-    }
-
     final override fun inflate(parent: ViewGroup): View {
         return createRecyclerView(parent).apply {
             recyclerView = this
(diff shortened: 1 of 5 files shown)


bento/src/main/java/com/yelp/android/bento/components/CarouselComponent.kt
```diff
@@ -69,17 +72,40 @@ open class CarouselComponentViewHolder : ComponentViewHolder<Unit?, CarouselView
...
+    lateinit var element: CarouselViewModel
     private var attachedPool = false
 
+    /**
+     * Saves the scroll position of the carousel to the view model so that it can be restored when
+     * the carousel recyclerView itself is recycled in the larger component list. Scroll position
+     * is saved when the scroll state of the carousel transitions to
+     * [RecyclerView.SCROLL_STATE_IDLE].
+     */
+    private fun saveScrollPosition() {
```
dwaxemberg: Can you move this to the end of the class to follow class structure convention?

bento/src/main/java/com/yelp/android/bento/components/CarouselComponent.kt
```diff
@@ -93,6 +119,12 @@ open class CarouselComponentViewHolder : ComponentViewHolder<Unit?, CarouselView
         } else {
             controller.addComponent(group)
         }
+
+        recyclerView.post {
```
dwaxemberg, line 111: Why is the `post` necessary? Can you add a comment about it?

bento/src/main/java/com/yelp/android/bento/components/CarouselComponent.kt
```diff
@@ -69,17 +72,40 @@ open class CarouselComponentViewHolder : ComponentViewHolder<Unit?, CarouselView
...
+     * Saves the scroll position of the carousel to the view model so that it can be restored when
+     * the carousel recyclerView itself is recycled in the larger component list. Scroll position
+     * is saved when the scroll state of the carousel transitions to
+     * [RecyclerView.SCROLL_STATE_IDLE].
+     */
+    private fun saveScrollPosition() {
+        (recyclerView.layoutManager as? LinearLayoutManager)?.apply {
+            element.scrollPosition = findFirstVisibleItemPosition()
+            element.scrollPositionOffset =
+                    recyclerView.getChildAt(0).left - recyclerView.paddingLeft
```
dwaxemberg: Is it possible for the recyclerview to be empty when this is called? (ie. cause an index out of bounds exception)

dwaxemberg (changes requested):

bento/src/main/java/com/yelp/android/bento/components/CarouselComponent.kt
argo49: I want to say it's not possible, since the only time we save the position is when the carousel scroll listener is triggered, and I think that's impossible to trigger on an empty recyclerview. I don't have any evidence of that though so I'll just add a test and a check to be safe.

dwaxemberg (approved):

PR #32, opened 2019-08-27
Update .idea files for AS 3.5

MatthewTPage:
I'm not sure how much of these .idea files were intentionally set, but assuming they were left as AS defaults like the rest of our repos at Yelp, this is what AS tries to change the .idea files to on AS 3.5, at least on my machine.

This is probably not backwards compatible with AS 3.4 and below.


dwaxemberg:
Can you update the readme to say version 3.5 instead of 3.3.2?

dwaxemberg (changes requested):
Update the readme to say we recommend Android Studio 3.5 instead of the current 3.3.2.

dwaxemberg (approved):

dwaxemberg:
merged in 5ee2548

Issue #30, opened 2019-08-20
there is some bug in ListComponent.onItemRangeInserted


eenie:
when I append 3 datas in ListComponent 
it just can show one new item 
and then I append 3 datas again
it show the last 2 item , the new data is gone
and repeat
btw, I try it in the demo

eenie:
the code is `  override fun onRefresh() {
        (componentController[1] as ListComponent<Nothing, String>).appendData((1..3).map { "List new element $it" })
        lay_swiper.isRefreshing = false
    }`

dwaxemberg:
Thanks for reporting this! 

The issue is `ListComponent.appendData()` isn't respecting the dividers when it notifies that items have been inserted. If you disable the dividers in the list using `toggleDivider(false)`, your example will work properly.

dwaxemberg:
Fixed by #33 

Bumping to `15.3.1` should solve your issue.

PR #33, opened 2019-08-29
Fixed appending data to ListComponent

dwaxemberg:
When a ListComponent has dividers the appendData method was not properly handling the divider offsets and only some of the items would be properly added. This change adds a check for the proper indices when there are dividers.

This addresses #30


bento/src/main/java/com/yelp/android/bento/components/ListComponent.java
```diff
@@ -79,8 +79,8 @@ public void setData(@NonNull List<T> data) {
      * @param data The new data list items to add.
      */
     public void appendData(@NonNull List<T> data) {
-        int oldSize = mData.size();
-        int sizeChange = data.size();
+        int oldSize = mShouldShowDivider ? getTotalSizeWithSeparators(mData.size()) : mData.size();
+        int sizeChange = mShouldShowDivider ? data.size() * 2 : data.size();
```
argo49, line 83: Why not also use `getTotalSizeWithSeparators` here? The last item shouldn't have a divider, so we still need to subtract 1, no?

dwaxemberg: We do not want to subtract 1 in this case. The difference here, is that we need to have a divider _before_ the new items we are inserting, otherwise there would be 2 items touching. 
For example, if we have a list with 2 items and we add 2 items we would have:
```
--item 1--   // index 0
--divider--  // index 1
--item 2--  // index 2

add(2)
--item 1--   // index 0
--divider--  // index 1
--item 2--  // index 2
--divider--  // index 3   <--- this divider is the extra one
--item 3--   // index 4
--divider--  // index 5
--item 4--  // index 6
```
Notice that we would need to call notify (2, 4) **not** (2, 3)

argo49 (approved):

dwaxemberg:
Fixes #30 #34

PR #38, opened 2019-10-08
Add check for -1 when reordering items and explicitly throwing detailed exception when rangedValueAt has an error.

wyattjsmith1:
We have a rare crash we can not repro in `getMovementFlags`. The leading theory is that `viewHolder.adapterPosition` is returning  `RecyclerView.NO_POSITION`, which it explicitly can. This is then being passed to `rangedValueAt`, where it is crashing.

Our solution is twofold:
1. Check if `viewHolder.adapterPosition == RecyclerView.NO_POSITION` and return if it does.
2. Add an explicit and detailed exception when `rangedValueAt` can't find an item to make debugging easier in the future.

Note: Because we can not repro this, we can not test the solution to ensure it fixes the problem. I have tested all of the screens in the Bento sample app, so I am confident that, though this may not fix the problem, it will not cause any harm.

### DIFF
--- a/bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
+++ b/bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
@@ -39,11 +39,11 @@ class ListItemTouchCallback(
             viewHolder: RecyclerView.ViewHolder
     ): Int {
         val currentIndex = viewHolder.adapterPosition
-        return if (currentIndex == RecyclerView.NO_POSITION ||
-                !canReorderItemAtIndex(currentIndex)) {
-            makeMovementFlags(0, 0)
-        } else {
+        return if (currentIndex != RecyclerView.NO_POSITION &&
+                canReorderItemAtIndex(currentIndex)) {
             makeMovementFlags(DRAG_FLAGS, 0)
+        } else {
+            makeMovementFlags(0, 0)
         }
     }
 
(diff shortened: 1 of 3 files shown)


bento/src/main/java/com/yelp/android/bento/utils/AccordionList.java
```diff
@@ -66,7 +66,13 @@ public T valueAt(int location) {
      */
     @NonNull
     public RangedValue<T> rangedValueAt(int location) {
-        return mRangeList.get(Collections.binarySearch(mRangeList, location));
+        int foundLocation = Collections.binarySearch(mRangeList, location);
+
+        if (foundLocation < 0) {
+            throw new ArrayIndexOutOfBoundsException("Could not find value at index: " + location +
+                    ". BinarySearch returned: " + foundLocation);
```
dwaxemberg: I think we should include the size of the list and maybe even the stored ranges...

bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
```diff
@@ -38,10 +38,12 @@ class ListItemTouchCallback(
...
-        return if (canReorderItemAtIndex(viewHolder.adapterPosition)) {
-            makeMovementFlags(DRAG_FLAGS, 0)
-        } else {
+        val currentIndex = viewHolder.adapterPosition
+        return if (currentIndex == RecyclerView.NO_POSITION ||
+                !canReorderItemAtIndex(currentIndex)) {
             makeMovementFlags(0, 0)
+        } else {
+            makeMovementFlags(DRAG_FLAGS, 0)
         }
```
dwaxemberg: I think we should invert the if and the else to have the success case first

dwaxemberg (changes requested):

bento/src/main/java/com/yelp/android/bento/core/ListItemTouchCallback.kt
dwaxemberg: Also, don't forget to update the CHANGELOG file

dwaxemberg (approved):

PR #41, opened 2019-12-10
Add a new NestedComponent

frostyshadows:
This component supports nesting an inner component inside an outer viewHolder, will be useful for adding borders to components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bento Releases
 
+## Version 15.3.3
+_2019-12-16_
+* New: Add a NestedComponent that can be used for nesting a given inner Component inside a given outer ViewHolder.
+
 ## Version 15.3.2
 _2019-10-08_
 * Fix: Fixed bug in ListItemTouchCallback causing crash when viewHolder.adapterPosition is -1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Bento Releases
 
-## Version 15.3.3
-_2019-12-16_
+## Version 15.4.0
+_2019-12-18_
 * New: Add a NestedComponent that can be used for nesting a given inner Component inside a given outer ViewHolder.
 
 ## Version 15.3.2
--- a/bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt
+++ b/bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt
@@ -2,6 +2,7 @@ package com.yelp.android.bento.components
 
 import android.view.View
 import android.view.ViewGroup
+import androidx.annotation.CallSuper
 import androidx.annotation.IdRes
 import androidx.annotation.LayoutRes
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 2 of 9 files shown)


bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt
```diff
@@ -0,0 +1,73 @@
...
+    abstract val recyclerViewId: Int
+
+    override fun inflate(parent: ViewGroup): View {
+        return parent.inflate<View>(outerLayout).apply {
+            val recyclerView: RecyclerView = findViewById(recyclerViewId)
+            controller = RecyclerViewComponentController(recyclerView, RecyclerView.VERTICAL)
+        }
+    }
+
+    override fun bind(presenter: Any?, element: NestedViewModel<T>) {
```
dwaxemberg: We should add an @CallSuper annotation to this function to guarantee subclasses will behave correctly.

bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt
```diff
@@ -0,0 +1,73 @@
...
+    override fun getPresenter(position: Int) = viewModel.innerComponent.getPresenter(position)
+
+    override fun getItem(position: Int) = viewModel
+
+    override fun getCount() = 1
+
+    override fun getHolderType(position: Int) = viewModel.outerComponentViewHolder
+}
+
+abstract class NestedOuterComponentViewHolder<T> : ComponentViewHolder<Any?, NestedViewModel<T>>() {
```
dwaxemberg: I think it would be beneficial to also have the presenter be a generic type in case a subclass cares about the presenter in the `bind` method.

bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt
```diff
@@ -0,0 +1,73 @@
...
+    abstract val outerLayout: Int
+
+    /**
+     * Id of recyclerView widget inside the outer layout. The recyclerView will be used to create
+     * the component controller.
+     */
+    @get:IdRes
+    abstract val recyclerViewId: Int
+
+    override fun inflate(parent: ViewGroup): View {
```
dwaxemberg, line 53: We should make this function final to preserve the proper behavior.

dwaxemberg (changes requested):
Looking good. I left a few minor suggestions

bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt
frostyshadows: I added a @CallSuper since it's possible subclasses will want to override `inflate()`, for example to do some `findViewByIds()`.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Bento Releases
 
+## Version 15.3.3
```
dwaxemberg: This change adds a new feature so it should be a minor version bump (we go by [semver](https://semver.org/)) ie. `15.4.0`

bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt
```diff
@@ -0,0 +1,77 @@
...
+     */
+    @get:IdRes
+    abstract val recyclerViewId: Int
+
+    @CallSuper
+    override fun inflate(parent: ViewGroup): View {
+        return parent.inflate<View>(outerLayout).apply {
+            val recyclerView: RecyclerView = findViewById(recyclerViewId)
+            controller = RecyclerViewComponentController(recyclerView, RecyclerView.VERTICAL)
+        }
```
dwaxemberg: We should use `also` here instead of `apply` since we are performing side-effects rather than acting on the object itself.

We can probably drop the `recyclerView` variable as well since we only use it once.

dwaxemberg (changes requested):

dwaxemberg (approved):

dwaxemberg:
merged in a1fd768

PR #44, opened 2020-01-27
Change NestedComponent to have separate presenter from its inner Component

victorslkim:
Currently, `NestedComponent`'s presenter is set to be inner `Component`'s presenter 

https://github.com/Yelp/bento/blob/48002fc6e1e920464364858e729d489a6b77a16c/bento/src/main/java/com/yelp/android/bento/components/NestedComponent.kt#L24

This makes it harder to re-use `NestedOuterComponentViewHolder`, which is `NestedComponent`'s `ComponentViewHolder`, since its presenter type must match with inner `Component`'s presenter type.

If we want to add multiple `NestedComponent` and the inner `Component`s have different presenter types, then we have to create multiple subclasses of `NestedOuterComponentViewHolder` and use different `NestedOuterComponentViewHolder`s for each `NestedComponent`. 

For example, if inner `Component`s have presenter type A, B, C, then we also have to create subclasses of `NestedOuterComponentViewHolder` with presenter type A, B, C.

By changing `NestedComponent` to have separate presenter from its inner `Component`, we can avoid creating multiple subclasses of `NestedOuterComponentViewHolder`, and make it easier to use `NestedComponent`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bento Releases
 
+## Version 15.6.0
+_2020-01-27_
+* Fix: Change NestedComponent to have separate presenter from its inner Component.
+
 ## Version 15.5.0
 _2020-01-16_
 * New: Add remove() and clear() functions to CarouselComponent.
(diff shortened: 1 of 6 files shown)


dwaxemberg (approved):
Can you also update the CHANGELOG file?

PR #46, opened 2020-02-04
Propagate CarouselComponent visibility to it's children

### LINKED ISSUE #45: Propagate Carousel's visibility change to its nested children
When a Carousel is scrolled away from view in a recycler view, it's nested components are not notified that they are becoming invisible.

When the carousel get's back into view, its nested component then don't get notified that they are becoming visible again.

This behavior prevents per card actions to be generated

redwarp:
Also add a visibility listener to Components

This changes addresses an issue with the CarouselComponent: It would not propagate it's visibility status to it's children.
With this change this is now fixed, and the children components that are currently on screen will get notified if they become visible or invisible.

Also, to avoid people having to subclass component to use visibility listener, adds a set of method to register and unregister listeners.

Fixes #45

### DIFF
--- a/bento/src/androidTest/java/com/yelp/android/bento/test/CarouselComponentViewHolderTest.kt
+++ b/bento/src/androidTest/java/com/yelp/android/bento/test/CarouselComponentViewHolderTest.kt
@@ -17,7 +17,7 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNull
 import org.junit.Test
 
-class CarouselComponentViewHolderTest: ComponentViewHolderTestCase<Unit?, CarouselViewModel>() {
+class CarouselComponentViewHolderTest : ComponentViewHolderTestCase<Unit?, CarouselViewModel>() {
 
     @Test
     fun providingRecycledPool_SharesPool() {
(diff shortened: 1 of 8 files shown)


bento/src/androidTest/java/com/yelp/android/bento/test/CarouselComponentViewHolderTest.kt
```diff
@@ -142,6 +142,28 @@ class CarouselComponentViewHolderTest: ComponentViewHolderTestCase<Unit?, Carous
         }
     }
 
+    @Test
+    fun scrollCarousel_withItemVisibilityListener_getsCalledForEveryItem(){
+        val group = ComponentGroup()
+        group.addAll((1..20).map { SimpleComponent<Unit>(TestComponentViewHolder::class.java) })
+        var visibleCount = 0
+        group.registerItemVisibilityListener { index, isVisible ->
+            if(isVisible) visibleCount++
```
dwaxemberg: nit: add space after `if`

dwaxemberg (approved):

bento/src/androidTest/java/com/yelp/android/bento/test/CarouselComponentViewHolderTest.kt
redwarp: Oh, I'm too used to auto-formatting hooks :-D Will do!

PR #48, opened 2020-02-07
feat: Add an isScrollable boolean flag to ComponentController

### LINKED ISSUE #36: Making recyclerview not scrollable is hard
Right now is not possible to have access to `LayoutManager` and make changes to its scrollable methods. We could expose scrolling options on the constructor or allow `RecyclerViewComponentController` to receive a `LayoutManager` making it more flexible to other use case.

I didn't went through all the controllers but I think either exposing `LayoutManager` properties or itself entirely are good options to increase flexibility for the recyclerview and decouple `RecyclerViewComponentController` from `LayoutManager`

redwarp:
The ability to actually change the scrollability will only work on RecyclerViews.
ListViews and ViewPager need to be subclassed to do so as well.

Solves #36

### DIFF
--- a/bento/src/main/java/com/yelp/android/bento/componentcontrollers/ListViewComponentController.kt
+++ b/bento/src/main/java/com/yelp/android/bento/componentcontrollers/ListViewComponentController.kt
@@ -46,6 +46,18 @@ class ListViewComponentController(val listView: ListView) :
         listView.setTag(R.id.bento_list_component_controller, this)
     }
 
+    /**
+     * The scrolling capacity of a [ListView] depends on the ListView itself. To make a
+     * ListView non scrollable, you must subclass the ListView itself.
+     */
+    @Suppress("UNUSED_PARAMETER")
+    override var isScrollable: Boolean
+        get() = listView.canScrollHorizontally(-1)
+                || listView.canScrollHorizontally(1)
+                || listView.canScrollVertically(-1)
+                || listView.canScrollVertically(1)
+        set(value) = Unit
+
     override fun get(index: Int) = components[index]
 
     override fun contains(component: Component) = component in components
--- a/bento/src/main/java/com/yelp/android/bento/componentcontrollers/ViewPagerComponentController.java
+++ b/bento/src/main/java/com/yelp/android/bento/componentcontrollers/ViewPagerComponentController.java
@@ -222,6 +222,25 @@ public class ViewPagerComponentController extends PagerAdapter implements Compon
         mComponentPageMap.remove(component);
     }
 
+    @Override
+    public boolean isScrollable() {
+        if (mViewPager != null) {
+            return mViewPager.isVerticalScrollBarEnabled()
+                    || mViewPager.isHorizontalScrollBarEnabled();
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * The scrolling capacity of a {@link ViewPager} depends on the ViewPager itself. To make a
+     * ViewPager non scrollable, you must subclass the ViewPager itself.
+     *
+     * @param isScrollable Ignored by Bento.
+     */
+    @Override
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Bento Releases
+## Unreleased version
+* New: Add an `isScrollable` boolean flag to ComponentController
```
dwaxemberg: Do we still need this?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/ListViewComponentController.kt
```diff
@@ -29,6 +29,7 @@ class ListViewComponentController(val listView: ListView) :
     override val span: Int get() = components.span
     override val size: Int get() = components.size
 
+
```
dwaxemberg: nit: remove extra line

bento/src/main/java/com/yelp/android/bento/componentcontrollers/ListViewComponentController.kt
```diff
@@ -46,6 +47,18 @@ class ListViewComponentController(val listView: ListView) :
...
+     * The scrolling capacity of a [ListView] depends on the ListView itself. To make a
+     * ListView non scrollable, you must subclass the ListView itself.
+     */
+    @Suppress("UNUSED_PARAMETER")
+    override var isScrollable: Boolean
+        get() = listView.canScrollHorizontally(-1)
+                || listView.canScrollHorizontally(1)
+                || listView.canScrollVertically(-1)
+                || listView.canScrollVertically(1)
+        set(value) = Unit
```
dwaxemberg, line 59: Hmm.. for this could we set the touch listener on the ListView and consume all the DRAG actions if this value is toggled?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/ViewPagerComponentController.java
```diff
@@ -222,6 +222,25 @@ public void destroyItem(@NonNull ViewGroup container, int position, @NonNull Obj
...
+    }
+
+    /**
+     * The scrolling capacity of a {@link ViewPager} depends on the ViewPager itself. To make a
+     * ViewPager non scrollable, you must subclass the ViewPager itself.
+     *
+     * @param isScrollable Ignored by Bento.
+     */
+    @Override
+    public void setScrollable(boolean isScrollable) {} // Do nothing.
```
dwaxemberg, line 242: Similar to ListView, could we use the touch listener to do this?

dwaxemberg (changes requested):

CHANGELOG.md
redwarp: We don't!

bento/src/main/java/com/yelp/android/bento/componentcontrollers/ListViewComponentController.kt
redwarp: So, touch listener seem to be inconsistent in how they work. Typically, it does not work for devices with keyboard, as they can always scroll.
One possibility could be to also intercept keyboard events?

redwarp: Also, apparently using the touch listener can be tricky: if one of the view inside the list view intercept the touch (like, a button), then the touch listener will be ignored, and a swipe will be possible.

dwaxemberg (approved):

bento/src/main/java/com/yelp/android/bento/componentcontrollers/ListViewComponentController.kt
redwarp: ++

Issue #49, opened 2020-02-07
Adopt conventional commits in our workflow


redwarp:
I would like this repo to adopt conventional commits as a way to enforce that our changelog is consistent and can be used to automatically generate changelogs when we are release new versions.

https://www.conventionalcommits.org/en/v1.0.0/

A quick overview:

> The commit contains the following structural elements, to communicate intent to the consumers of your library:
> 
> fix: a commit of the type fix patches a bug in your codebase (this correlates with PATCH in semantic versioning).
> feat: a commit of the type feat introduces a new feature to the codebase (this correlates with MINOR in semantic versioning).
> BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with MAJOR in semantic versioning). A BREAKING CHANGE can be part of commits of any type.
> types other than fix: and feat: are allowed, for example @commitlint/config-conventional (based on the the Angular convention) recommends build:, chore:, ci:, docs:, style:, refactor:, perf:, test:, and others.
> footers other than BREAKING CHANGE: <description> may be provided and follow a convention similar to git trailer format.

The benefits of this kind of format is that thanks to its structure, tooling exist around it to automatically generate change logs in a readable and structured form based on the commit messages. See https://www.conventionalcommits.org/en/v1.0.0/#tooling-for-conventional-commits

Also, it is possible to add some github actions like https://github.com/outillage/commitsar directly to the project to make sure that every commit follows the standard.

PR #51, opened 2020-02-19
chore: Setup conventional commits (#49)

redwarp:
Setup commitsar as a github action, and prepare for usage of standard-version as future release tool.

Let me what you think, and maybe try to create some releases (don't push :-P) to see if everything seems to work, or if we should improve!

As you can see, this change ads the github action that makes an additional check for every pull request, to check that the commit follows standards.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -1,0 +1,11 @@
+name: Pull request
+on: pull_request
+
+jobs:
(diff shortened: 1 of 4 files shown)


redwarp:
Odd, commitsar did not run.

redwarp:
I think it is not running because the branch is currently on another repo. I'm canceling this, and will create a PR from this repository.

PR #52, opened 2020-02-19
chore: Setup conventional commits (#49)

### LINKED ISSUE #49: Adopt conventional commits in our workflow

redwarp:
Setup commitsar as a github action, and prepare for usage of standard-version as future release tool.

Let's check if commitsar can run if I'm creating the PR from a branch on this repo, and not the fork.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -1,0 +1,11 @@
+name: Pull request
+on: pull_request
+
+jobs:
+  commitsar:
+    runs-on: ubuntu-latest
+    name: Verify commit messages
+    steps:
+      - uses: actions/checkout@v1
+      - name: Run commitsar
+        uses: docker://commitsar/commitsar
--- a/README.md
+++ b/README.md
@@ -107,13 +107,33 @@ Repeat as necessary. The project should build from within Android Studio and if 
 
 You can also run `./gradlew publishToMavenLocal` to publish the package to your local maven repo on your machine. From there you can add `mavenLocal()` before other maven repositories. That way your project can load in the version of Bento you're working on.
 
-We follow [semver](https://semver.org/) so the version number in `GlobalDependencies.kt` should be updated to reflect the changes.
-
-Add a summary of your change to the `CHANGELOG.md` file.
+We follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) specification, and you should write what your changes are about in a clear commit message.
 
 #### 5. Push your changes and open a pull request
 
-Push your branch with the new commits to your cloned repo, then open a pull request through GitHub. Once Travis CI gives it a green light along with a member of our team, it will be merged and the new version will be deployed to Maven Central.
+First, squash your commits if you have multiple, and make sure your commit message follows the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) convention.
+Push your branch with the new commits to your cloned repo, then open a pull request through GitHub. Once Travis CI gives it a green light along with a member of our team, it will be merged.
+
+If your reviewer asks for some changes, address the issue and then make sure to amend your previous commit - there should be only one commit per pull request. Push your changes, repeat until you get a green light.
+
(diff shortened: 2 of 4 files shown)


redwarp:
Yep, you can see the action from here.

README.md
```diff
@@ -107,13 +107,33 @@ Repeat as necessary. The project should build from within Android Studio and if
...
+## Making a release
+We switched to using [standard-version](https://github.com/conventional-changelog/standard-version) to handle our releases. standard-version is a utility for versioning using [semver](https://semver.org/) and CHANGELOG generation powered by Conventional Commits.
+
+If you are one of the members and must release a new version, then check its documentation. You will need to install `npm` if you don't have it. (We recommand using `homebrew` to do so.)
+
+With `npm` install, install standard-version globally: `npm i -g standard-version`
+
+From master, run `standard-version`.
+
+`standard-version` will take care of
```
argo49: nit: remove "take care of" so we don't need to change the tense of the verbs in the list

README.md
```diff
@@ -107,13 +107,33 @@ Repeat as necessary. The project should build from within Android Studio and if
...
+* Detect the previous version tag (based on git tags)
+* Infer the next version number, based on previous standard commit messages. A breaking change will trigger a major bump, a new feature will trigger a minor bump, and a fix will trigger a patch bump.
+* Update the CHANGELOG.md file with the commit messages, as well as the [GlobalDependency.kt](buildSrc/src/main/java/com/yelp/gradle/bento/GlobalDependencies.kt) file with the new version number.
+* Create a new commit with these change, and tag it with the version number.
+
+The rest has to be done manually: check that the commit looks good (proper changelog, version number, tag), then push both commit and tags to master.
+
+Publishing to maven is also done manually for now. Check with other members about how to proceed.
```
argo49: Might be a good idea to mention we have an internal wiki page on deploying to maven.

argo49 (approved):
This is pretty cool!

README.md
redwarp: Good point

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,11 @@
+name: Pull request
+on: pull_request
+
+jobs:
+  commitsar:
+    runs-on: ubuntu-latest
```
dwaxemberg, line 6: Is this safe? Should we pin to an LTS version?

dwaxemberg (approved):

.github/workflows/pull_requests.yml
redwarp: @dwaxemberg I checked, and apparently that is how github does action 🤷‍♂

PR #55, opened 2020-02-21
docs: Fixed typo in wiki example description

coltin:


redwarp:
So @coltin  we decided to embrace conventional commits ;-) So if you amend your commit message, should be good

coltin:
@redwarp Never heard of that before, neat! Fixed

redwarp:
@coltin techically, this should not be `fix:` but `doc:`

redwarp:
So it should be reworded: a "fix" will show up in the  changelog.

redwarp (commented):
Good to go

PR #72, opened 2021-02-10
fix: Fix crash in ComponentVisibilityListener

### LINKED ISSUE #71: Crash in ComponentVisibilityListener
In the ComponentVisibilityListener, the method that tries to notify nested components does not actually verifies that these components are visible.

https://github.com/Yelp/bento/blob/c459cddf3a842da7b77000b60cb575457c596523/bento/src/main/java/com/yelp/android/bento/core/ComponentVisibilityListener.kt#L152

We need to add a guard to firstVisible and lastVisible to verify they are not a NO_POSITION

redwarp:
Actually verifies that the item positions are different to NO_POSITION, which will prevent the index out of bounds exception.

Fixes #71


dwaxemberg (approved):
I think we should change it to an OR even though they _should_ be the same. Just to be safe

redwarp:
Updated to use OR

PR #74, opened 2021-05-27
chore: Upgrade Guava to 28.1

carolkng:
Currently Bento is using an older version of Guava (25.0) which results in dependency conflicts when trying to use the most up-to-date version of [Google's accessibility test framework](https://github.com/google/Accessibility-Test-Framework-for-Android). 

Ran bento tests and didn't see anything out of the ordinary.


redwarp (approved):

redwarp:
Let's amend the commit message to something like "fix: Bump guava" and then I'll make a new bento release.

PR #78, opened 2021-05-27
chore: Setup workflow to replace travis ci

redwarp:
Maintaining travis seems a bit annoying, as we need to specify build version, accept license, etc... All of that that seem to work automatically with the github actions.
The idea would be to remove the travis run if this gets merged.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -3,9 +3,60 @@ on: pull_request
 
 jobs:
   commitsar:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest]
     name: Verify commit messages
     steps:
       - uses: actions/checkout@v1
       - name: Run commitsar
         uses: docker://commitsar/commitsar
+  checks:
+    runs-on: [ubuntu-latest]
+    name: Run checks
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - name: Cache Gradle Folders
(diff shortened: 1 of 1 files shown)


redwarp:
Failing is normal here, it will work once the update of gradle and android tools is merged.

MatthewTPage (approved):

MatthewTPage:
might be good to wait for @dwaxemberg on this one as well

redwarp:
Ping @dwaxemberg

.github/workflows/pull_requests.yml
```diff
@@ -3,9 +3,60 @@ on: pull_request
...
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches/
+            ~/.gradle/wrapper/
+          key: cache-gradle-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}
+          restore-keys: |
+            cache-gradle-
+      - name: Run publishToMavenLocal
+        run: ./gradlew publishToMavenLocal
```
dwaxemberg, line 62: What is this task used for?

dwaxemberg (approved):
Looks good

.github/workflows/pull_requests.yml
redwarp: That's a good question. I added it because it's part of the original travis list of tasks being run, and I imagine it's there to make sure there is no error during the construction of artifacts.
But it's debatable if this would catch any error anyway.

PR #79, opened 2021-06-25
Added option to inflate views on a background thread.

martinpdd8:
### This change: 
Tried to use `AsyncLayoutInflater` to inflate views off of the main thread. `AsyncInflationBridge` acts as a ...bridge between `RecyclerViewComponentController` and the underlying `RecyclerView.Adapter`, intercepting the `addComponent()` calls and inflating components' views on a background thread. As a bonus, I offload some of the reflective `constructViewHolder` calls to a background thread and store 'em in a cache that `RecyclerViewComponentController` can use later.

### Does it help? 

Preliminary tests on a Pixel 4 XL (real device) show a ~25% improvement in render timings for the first few frames of every fragment / activity.

image

### DIFF
--- a/bento-sample-app/src/androidTest/java/com/yelp/android/bentosampleapp/RecyclerViewActivityTest.kt
+++ b/bento-sample-app/src/androidTest/java/com/yelp/android/bentosampleapp/RecyclerViewActivityTest.kt
@@ -1,0 +1,31 @@
+package com.yelp.android.bentosampleapp
+
+import androidx.test.espresso.Espresso.onView
+import androidx.test.espresso.action.ViewActions
+import androidx.test.espresso.assertion.ViewAssertions
+import androidx.test.espresso.intent.rule.IntentsTestRule
+import androidx.test.espresso.matcher.ViewMatchers
+import androidx.test.espresso.matcher.ViewMatchers.withId
+import com.yelp.android.bento.utils.BentoSettings
+import com.yelp.android.bentosampleapp.rules.BeforeRule
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.RuleChain
+
+class RecyclerViewActivityTest {
+
+    private val beforeRule = BeforeRule {
+        BentoSettings.asyncInflationEnabled = true
+    }
+    private val intentsTestRule = IntentsTestRule(RecyclerViewActivity::class.java)
+
+    @get:Rule
+    val rule: RuleChain = RuleChain.outerRule(beforeRule).around(intentsTestRule)
+
+    @Test
+    fun withAsyncInflationEnabled_scrollingDownDoesntCrash_andDisplaysItems() {
+        onView(withId(R.id.recyclerView)).perform(ViewActions.swipeUp())
+        onView(ViewMatchers.withText("List element 20"))
+                .check(ViewAssertions.matches(ViewMatchers.isDisplayed()))
+    }
+}
--- a/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
+++ b/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
@@ -12,6 +12,7 @@ import androidx.recyclerview.widget.RecyclerView.OnScrollListener;
 import androidx.recyclerview.widget.RecyclerView.Orientation;
 import androidx.recyclerview.widget.RecyclerView.ViewHolder;
 import com.google.common.collect.HashBiMap;
+import com.yelp.android.bento.core.AsyncInflationBridge;
 import com.yelp.android.bento.core.BentoLayoutManager;
 import com.yelp.android.bento.core.Component;
 import com.yelp.android.bento.core.Component.ComponentDataObserver;
@@ -26,6 +27,7 @@ import com.yelp.android.bento.core.ListItemTouchCallback;
 import com.yelp.android.bento.core.OnItemMovedPositionListener;
 import com.yelp.android.bento.utils.AccordionList.Range;
 import com.yelp.android.bento.utils.AccordionList.RangedValue;
+import com.yelp.android.bento.utils.BentoSettings;
 import com.yelp.android.bento.utils.Sequenceable;
 import java.util.Collection;
 import java.util.HashMap;
@@ -39,6 +41,7 @@ import org.jetbrains.annotations.Nullable;
 public class RecyclerViewComponentController
         implements ComponentController, OnItemMovedPositionListener {
 
+    private static final int INVALID_COMPONENT_INDEX = -1;
     private final RecyclerView.Adapter<ViewHolderWrapper> mRecyclerViewAdapter;
     private final ComponentGroup mComponentGroup;
     private final Map<Component, Set<Class<? extends ComponentViewHolder>>>
@@ -53,26 +56,63 @@ public class RecyclerViewComponentController
     private BentoLayoutManager mLayoutManager;
     private LinearSmoothScroller mSmoothScroller;
     private ItemTouchHelper mItemTouchHelper;
+    public AsyncInflationBridge mAsyncInflationBridge;
 
     @RecyclerView.Orientation private int mOrientation;
+
+    private final boolean mAsyncInflationEnabled;
 
     /**
      * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
      * the {@link RecyclerView}. In order to make the lanes, this component controller will set the
-     * {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do it manually.
+     * {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do it manually. Async
+     * inflation will be enabled based on the global toggle in {@link BentoSettings} which is
+     * disabled by default. When enabled, views will be inflated on a background thread.
      */
     public RecyclerViewComponentController(@NonNull RecyclerView recyclerView) {
-        this(recyclerView, RecyclerView.VERTICAL);
+        this(recyclerView, RecyclerView.VERTICAL, BentoSettings.getAsyncInflationEnabled());
+    }
+
+    /**
+     * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
+     * the {@link RecyclerView}. Passing true for 'asyncInflationEnabled' will enable {@link
+     * AsyncInflationBridge} for this controller, which inflates component's views on a background
+     * thread. This flag will override the global flag in {@link BentoSettings}.
+     */
+    public RecyclerViewComponentController(
+            @NonNull RecyclerView recyclerView, boolean asyncInflationEnabled) {
+        this(recyclerView, RecyclerView.VERTICAL, asyncInflationEnabled);
     }
 
     /**
      * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
      * the {@link RecyclerView}. In order to make the lanes (columns / rows), this component
      * controller will set the {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do
-     * it manually.
+     * it manually. Async inflation will be enabled based on the global toggle in {@link
+     * BentoSettings} which is disabled by default. When enabled, views will be inflated on a
+     * background thread.
      */
     public RecyclerViewComponentController(
             @NonNull RecyclerView recyclerView, @Orientation int orientation) {
+        this(recyclerView, orientation, BentoSettings.getAsyncInflationEnabled());
+    }
+
+    /**
+     * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
+     * the {@link RecyclerView}. In order to make the lanes (columns / rows), this component
+     * controller will set the {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do
+     * it manually. Passing true for 'asyncInflationEnabled' will enable {@link
+     * AsyncInflationBridge} for this controller, which inflates component's views on a background
+     * thread. This flag will override the global flag in {@link BentoSettings}.
+     */
+    public RecyclerViewComponentController(
+            @NonNull RecyclerView recyclerView,
+            @Orientation int orientation,
+            boolean asyncInflationEnabled) {
+        mAsyncInflationEnabled = asyncInflationEnabled;
+        if (asyncInflationEnabled) {
+            mAsyncInflationBridge = new AsyncInflationBridge(recyclerView);
+        }
         mOrientation = orientation;
         mRecyclerViewAdapter = new RecyclerViewAdapter();
         mComponentGroup = new ComponentGroup();
@@ -179,18 +219,32 @@ public class RecyclerViewComponentController
     @NonNull
     @Override
     public RecyclerViewComponentController addComponent(@NonNull Component component) {
-        mComponentGroup.addComponent(component);
-        shareViewPool(component);
-        mComponentVisibilityListener.onComponentAdded(component);
+        if (!mAsyncInflationEnabled) {
+            addComponentInternal(component);
+        } else {
+            mAsyncInflationBridge.asyncInflateViewsForComponent(
+                    component,
+                    () -> {
+                        addComponentInternal(component);
+                        return null;
+                    });
+        }
         return this;
     }
 
     @NonNull
     @Override
     public ComponentController addComponent(@NonNull ComponentGroup componentGroup) {
-        mComponentGroup.addComponent(componentGroup);
-        shareViewPool(componentGroup);
-        mComponentVisibilityListener.onComponentAdded(componentGroup);
+        if (!mAsyncInflationEnabled) {
+            addComponentInternal(componentGroup);
+        } else {
+            mAsyncInflationBridge.asyncInflateViewsForComponent(
+                    componentGroup,
+                    () -> {
+                        addComponentInternal(componentGroup);
+                        return null;
+                    });
+        }
         return this;
     }
 
@@ -198,18 +252,32 @@ public class RecyclerViewComponentController
     @Override
     public RecyclerViewComponentController addComponent(
             int index, @NonNull final Component component) {
-        mComponentGroup.addComponent(index, component);
-        shareViewPool(component);
-        mComponentVisibilityListener.onComponentAdded(component);
+        if (!mAsyncInflationEnabled) {
+            addComponentInternal(index, component);
+        } else {
+            mAsyncInflationBridge.asyncInflateViewsForComponent(
+                    component,
+                    () -> {
+                        addComponentInternal(index, component);
+                        return null;
+                    });
+        }
         return this;
     }
 
     @NonNull
     @Override
     public ComponentController addComponent(int index, @NonNull ComponentGroup componentGroup) {
-        mComponentGroup.addComponent(index, componentGroup);
-        shareViewPool(componentGroup);
-        mComponentVisibilityListener.onComponentAdded(componentGroup);
+        if (!mAsyncInflationEnabled) {
+            addComponentInternal(index, componentGroup);
+        } else {
+            mAsyncInflationBridge.asyncInflateViewsForComponent(
+                    componentGroup,
+                    () -> {
+                        addComponentInternal(index, componentGroup);
+                        return null;
+                    });
+        }
         return this;
     }
 
@@ -217,10 +285,23 @@ public class RecyclerViewComponentController
     @Override
     public RecyclerViewComponentController addAll(
             @NonNull Collection<? extends Component> components) {
-        mComponentGroup.addAll(components);
-        for (Component component : components) {
-            shareViewPool(component);
-            mComponentVisibilityListener.onComponentAdded(component);
+        if (!mAsyncInflationEnabled) {
+            mComponentGroup.addAll(components);
+            for (Component component : components) {
+                shareViewPool(component);
+                mComponentVisibilityListener.onComponentAdded(component);
+            }
+        } else {
+            for (Component component : components) {
+                mAsyncInflationBridge.asyncInflateViewsForComponent(
+                        component,
+                        () -> {
+                            addComponentInternal(component);
+                            shareViewPool(component);
+                            mComponentVisibilityListener.onComponentAdded(component);
+                            return null;
+                        });
+            }
         }
         return this;
     }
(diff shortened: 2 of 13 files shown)


bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -53,16 +54,38 @@
...
      * {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do it manually.
      */
     public RecyclerViewComponentController(@NonNull RecyclerView recyclerView) {
-        this(recyclerView, RecyclerView.VERTICAL);
+        this(recyclerView, RecyclerView.VERTICAL, true);
+    }
+
+    /**
+     * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
+     * the {@link RecyclerView} with async view inflation enabled.
```
dwaxemberg: Async inflation is passed in as a param right? We should update the comment.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -53,16 +54,38 @@
...
+    private int mOrientation;
+
+    private final boolean mAsyncInflationEnabled;
 
     /**
      * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
      * the {@link RecyclerView}. In order to make the lanes, this component controller will set the
      * {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do it manually.
      */
     public RecyclerViewComponentController(@NonNull RecyclerView recyclerView) {
```
dwaxemberg, line 72: The comment for this should probably include something about async inflation.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -179,37 +207,73 @@ public Range rangeOf(@NonNull Component component) {
...
+            shareViewPool(component);
+            mComponentVisibilityListener.onComponentAdded(component);
+            return this;
+        }
+        mAsyncInflationBridge.asyncInflateViewsForComponent(component, () -> {
+            mComponentGroup.addComponent(component);
+            shareViewPool(component);
+            mComponentVisibilityListener.onComponentAdded(component);
+            return null;
+        });
```
dwaxemberg: The block of code in the if and in the lambda are almost identical. I think we could pull it out into a helper function to reduce some duplication. (same for other `add` functions)

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -458,7 +522,7 @@ private void shareViewPool(@NonNull Component component) {
      *
      * @throws RuntimeException if the specified view holder type could not be instantiated.
      */
-    private ComponentViewHolder constructViewHolder(
+    public static ComponentViewHolder constructViewHolder(
```
dwaxemberg, line 562: Could this be package-protected instead of public? I don't think we want to expose this to consumers.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,91 @@
...
+        private val dispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
+    }
+
+    private val viewHolderMap = mutableMapOf<Class<out ComponentViewHolder<*, *>>,
+            MutableList<ComponentViewHolder<*, *>>>()
+    private val viewMap = mutableMapOf<ComponentViewHolder<*, *>, View?>()
+
+    /**
+     * Inflates the views and creates the view holder for [component] on a background thread,
+     * getting them nice and ready for RecyclerViewComponentController to use them. The
```
dwaxemberg: Trailing "the"

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,91 @@
...
+     */
+    fun asyncInflateViewsForComponent(component: Component, addComponentCallback: () -> Unit) {
+        GlobalScope.launch(dispatcher) {
```
dwaxemberg: I don't think we should use the GlobalScope for this. We probably want to pass in a lifecycle scope so we aren't inflating views after the activity/fragment is closed

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,91 @@
...
+            val deferredInflations = (0 until component.count).map { i ->
+                async {
+                    val viewHolderType = component.getHolderType(i)
+                    val viewHolder: ComponentViewHolder<*, *> = constructViewHolder(viewHolderType)
+                    addViewHolder(viewHolder, viewHolderType)
+                    val pair = BentoAsyncLayoutInflater.inflate(viewHolder, recyclerView)
```
dwaxemberg: You could deconstruct this so it's easier to read: `val (_, view) = Bento...`

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,91 @@
...
+internal class AsyncInflationBridge(val recyclerView: RecyclerView) {
+
+    private companion object {
+        private val lock = Mutex()
+        private val dispatcher = Executors.newSingleThreadExecutor().asCoroutineDispatcher()
+    }
+
+    private val viewHolderMap = mutableMapOf<Class<out ComponentViewHolder<*, *>>,
+            MutableList<ComponentViewHolder<*, *>>>()
+    private val viewMap = mutableMapOf<ComponentViewHolder<*, *>, View?>()
```
dwaxemberg: I think we need to use thread safe maps here

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,91 @@
...
+                    // TODO: handle crash if component is already added.
+                }
+            }
+        }
+    }
+
+    /**
+     * Retrieves a previously async inflated view if there is one.
+     */
+    @Synchronized
```
dwaxemberg: Why synchronize here?

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,91 @@
...
+    fun getView(viewHolder: ComponentViewHolder<*, *>) = viewMap[viewHolder]
+
+    /**
+     * Retrieves a previously created view holder of
```
dwaxemberg: of... ?

bento/src/main/java/com/yelp/android/bento/core/BentoAsyncLayoutInflater.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.withContext
+import java.util.concurrent.Executors
+
+
+private const val TAG = "BentoAsyncInflater"
+
+/**
+ * Inflates a componentViewHolder's views on one of ten lucky background threads.
```
dwaxemberg, line 15: lol

dwaxemberg (commented):
Overall this is looking great!

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
martinpdd8: @dwaxemberg It's used in two different packages so it has to be public. The other util methods are public too, such as `MathUtils.lcm` so I'm not sure how much it matters?

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: I started with that but since we added the mutex access should be mostly synchronous anyway... unless I'm missing something.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: This is required to pull views in order from the main thread. We could remove this and make `viewMap` the concurrent version instead, if you like?

bento/src/main/java/com/yelp/android/bento/utils/BentoAsyncInflationFlag.kt
```diff
@@ -0,0 +1,11 @@
+package com.yelp.android.bento.utils
+
+/**
+ * Global toggle for the async inflation feature in
+ * [com.yelp.android.bento.componentcontrollers.RecyclerViewComponentController]. When enabled,
+ * all RecyclerViewComponentControllers will try to inflate their component's views on a background
+ * thread. It's disabled by default.
+ */
+object BentoAsyncInflationFlag {
```
redwarp: If this a pattern we intend to keep, I would rather rename this class to BentoSettings, and rename `enabled` to `asyncInflationEnabled` so it can be used for other bento wide settings.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
redwarp: I still cannot understand why it would help, or do anything. We only synchronize `getView`, but getView is only called from the `RecyclerViewComponentController` class in the `onCreateViewHolder`. And I think this gets called by the RecyclerView and the GapWorker, but on the main thread. So I'm lost.

I'm scratching my head but I don't get it. I would feel more confident with a concurrent map I think, unless I finally understand why this work, or you can explain it.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private fun addViewHolder(
+            viewHolder: ComponentViewHolder<*, *>,
+            viewHolderType: Class<out ComponentViewHolder<*, *>>
+    ) {
+        var viewHolders = viewHolderMap[viewHolderType]
+        if (viewHolders == null) {
+            viewHolders = mutableListOf()
+        }
```
redwarp: More kotliny:

```
val viewHolders = viewHolderMap.getOrPut(viewHolderType){ mutableListOf() }
```
And as we said through slack, a concurrent list would probably be best

redwarp: Or at last, synchronize read/write operation in the list.

redwarp: So maybe this is sufficient: 

```
fun getViewHolder(
    viewHolderType: Any? // Class<out ComponentViewHolder<*, *>> doesn't work with Java.
): ComponentViewHolder<*, *>? {
    if (viewHolderMap.containsKey(viewHolderType)) {
        viewHolderMap[viewHolderType]?.let { list ->
            if (list.isNotEmpty()) {
                synchronized(this) {
                    return list.removeAt(0)
                }
            }
        }
    }
    return null
}

private fun addViewHolder(
    viewHolder: ComponentViewHolder<*, *>,
    viewHolderType: Class<out ComponentViewHolder<*, *>>
) {
    val viewHolders = viewHolderMap.getOrPut(viewHolderType) {
        mutableListOf()
    }
    synchronized(this) {
        viewHolders.add(viewHolder)
    }
    viewHolderMap[viewHolderType] = viewHolders
}
```

martinpdd8: I'm using a `ConcurrentLinkedDeque` for the list. I'll update it to use `getOrPut` too 👍

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: As we agreed on Slack, although there isn't much risk of a concurrent mishap for that map, it's still safer to use a `ConcurrentHashMap`. 👍

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: I made 'em concurrent.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -53,26 +56,64 @@
...
 
     /**
      * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
      * the {@link RecyclerView}. In order to make the lanes (columns / rows), this component
      * controller will set the {@link RecyclerView}'s {@link RecyclerView.LayoutManager}. Do not do
-     * it manually.
+     * it manually. Async inflation will be enabled based on the global toggle in
+     *
+     * @link BentoAsyncInflationFlag} which is disabled by default. When enabled, views will be
+     *     inflated on a background thread.
```
dwaxemberg: I think there's a missing '{' and an extra line

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -334,6 +416,20 @@ public void onItemPickedUp(ComponentViewHolder viewHolder) {
         }
     }
 
+    private void addComponentInternal(int index, @NonNull final Component component) {
+        if (index == -1) {
```
dwaxemberg: I think we should compare with INVALID_COMPONENT_INDEX instead of the literal here

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * This acts as a bridge between RecyclerViewComponentController and the underlying
+ * RecyclerView.Adapter. It intercepts the addComponent() calls and inflates views with
+ * [BentoAsyncLayoutInflater]. Components are added to the RecyclerView when the inflation
+ * is finished.
+ */
+internal class AsyncInflationBridge(val recyclerView: RecyclerView) {
+
+    private val context = recyclerView.context
```
dwaxemberg: Do we need this to be a val? I think we only use it once

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,116 @@
...
+                }
+            }
+            lock.withLock {
+                inflations.awaitAll()
+            }
+            withContext(Dispatchers.Main) {
+                try {
+                    addComponentCallback()
+                } catch (e: IllegalArgumentException) {
+                    // TODO: handle crash if component is already added.
```
dwaxemberg: Should we do this before pushing this PR?

buildSrc/src/main/java/com/yelp/gradle/bento/GlobalDependencies.kt
```diff
@@ -5,7 +5,7 @@ import org.gradle.api.artifacts.repositories.MavenArtifactRepository
 
 object Publishing {
     const val GROUP = "com.yelp.android"
-    const val VERSION = "18.0.2"
+    const val VERSION = "18.1.0-SNAPSHOT"
```
dwaxemberg: Just a reminder to change this when ready :)

dwaxemberg (approved):

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: If I pull it inside the `coroutineScope` block I'll have to cast it a second time to know it's a `LifecycleOwner`. I'm not sure that'd be neater than keeping the val as is. What do you think @dwaxemberg ?

dwaxemberg: Ah gotcha. This is fine then.

bento-sample-app/src/androidTest/java/com/yelp/android/bentosampleapp/RecyclerViewActivityTest.kt
```diff
@@ -0,0 +1,26 @@
...
+
+class RecyclerViewActivityTest {
+
+    @get:Rule
+    internal val intentsTestRule = BeforeActivityIntentsTestRule(RecyclerViewActivity::class.java) {
+        BentoSettings.asyncInflationEnabled = true
+    }
+
+    @Test
+    fun onData_atPosition2_clickOpensListViewActivity() {
```
dwaxemberg: nit: rename this test since it doesn't do what it says

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,163 @@
...
+        component: Component,
+        addComponentCallback: () -> Unit
+    ) {
+        try {
+            val timeAddComponentRequested = componentsAddedMap[component]
+            val timeRemoveComponentRequested = componentsRemovedMap[component]
+            if (timeRemoveComponentRequested != null && timeAddComponentRequested != null) {
+                if (timeAddComponentRequested > timeRemoveComponentRequested) {
+                    addComponentCallback()
+                } // else { The component was removed while we've been inflating. }
```
dwaxemberg: Just to make sure I understand what's happening here: This is the case where a component was added, we started inflating its views, it was removed, then added again, and then inflation finished for it so we are firing the callback?

dwaxemberg (approved):

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: Almost. This only checks that we aren't adding a component that was removed while we were inflating its views.

bento-sample-app/src/androidTest/java/com/yelp/android/bentosampleapp/rules/BeforeActivityIntentsTestRule.kt
```diff
@@ -0,0 +1,19 @@
+package com.yelp.android.bentosampleapp.rules
+
+import android.app.Activity
+import androidx.test.espresso.intent.rule.IntentsTestRule
+
+/**
+ * A simple [IntentsTestRule] that lets you run a block of code
+ * before the activity is launched.
+ */
+internal class BeforeActivityIntentsTestRule<T : Activity>(
```
redwarp: Neat:

It would be more generic to have something like a
```
class BeforeRule(val beforeCallback: () -> Unit) : ExternalResource() {

    override fun before() {
        beforeCallback()
    }
}
```
and then in the test class do 
```
private val beforeRule = BeforeRule {
    BentoSettings.asyncInflationEnabled = true
}
private val intentsTestRule = IntentsTestRule(RecyclerViewActivity::class.java)

@get:Rule
val rule: RuleChain = RuleChain.outerRule(beforeRule).around(intentsTestRule)
```
but... it's alright

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -334,6 +426,20 @@ public void onItemPickedUp(ComponentViewHolder viewHolder) {
...
+            mComponentGroup.addComponent(component);
+        } else {
+            mComponentGroup.addComponent(index, component);
+        }
+        shareViewPool(component);
+        mComponentVisibilityListener.onComponentAdded(component);
+    }
+
+    private void addComponentInternal(@NonNull final Component component) {
+        addComponentInternal(INVALID_COMPONENT_INDEX, component);
```
redwarp, line 435: How about `addComponentInternal(mComponentGroup.getSize(), component);`?

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -0,0 +1,163 @@
...
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
+
+/**
+ * This acts as a bridge between RecyclerViewComponentController and the underlying
+ * RecyclerView.Adapter. It intercepts the addComponent() calls and inflates views with
+ * [BentoAsyncLayoutInflater]. Components are added to the RecyclerView when the inflation
+ * is finished.
+ */
+internal class AsyncInflationBridge(
```
redwarp: If you write it like `internal class AsyncInflationBridge @JvmOverloads constructor(`, then in the java code, you would be able to simply call `AsyncInflationBridge(recyclerView)`

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
redwarp: We could also hold a map of the inflation jobs, like

```
private val inflationJobs = ConcurrentHashMap<Component, Job>()
```

and populate it like 
```
fun asyncInflateViewsForComponent(component: Component, addComponentCallback: () -> Unit) {
    componentsAddedMap[component] = System.nanoTime()
    inflationJobs[component] = coroutineScope.launch(defaultBridgeDispatcher) {
        val inflations = (0 until component.count).map { i ->

(...)
```

And in the trackComponentRemoval, do
```
fun trackComponentRemoval(component: Component) {
    inflationJobs[component]?.cancel()
}
```

But it's definitely an improvement we can keep for later.

redwarp (approved):
Nothing blocking, a few suggestions there and there.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: Thanks!

martinpdd8:
@redwarp It won't let me reply to the the last comment but it's _way_ better than what I had so I applied those changes. Thanks!

PR #83, opened 2021-08-06
Fixed a memory leak in the async inflation feature.

martinpdd8:
We were using whatever `lifecycleScope` `context` provided which lead to a memory leak when Bento was used in Fragments. The lifecycleScope was tied to the parent activity's lifecycle instead of the fragment and thus, unused inflated views would stick around in memory.

The solution I went with was to abandon the lifecycleScope and use a SupervisorJob that we clear ourselves. `findViewTreeLifecycleOwner()` finds the Fragment lifecycle owner if there is one, which we can then add a listener to to get notified about the destroy event. Since some RecyclerView's are created inside ViewHolder's, they aren't attached yet at the time this code runs. To get around that, I used a `addOnAttachStateChangeListener`. If no fragment view lifecycle is found, we fallback to the activity's lifecycleScope.

### DIFF
--- a/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
+++ b/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
@@ -343,6 +343,9 @@ public class RecyclerViewComponentController
         mComponentGroup.clear();
         removeVisibilityListeners();
         addVisibilityListeners();
+        if (mAsyncInflationEnabled) {
+            mAsyncInflationBridge.cancelAllInflationJobs();
+        }
     }
 
     @Override
--- a/bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
+++ b/bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
@@ -1,25 +1,27 @@
 package com.yelp.android.bento.core
 
 import android.view.View
-import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleObserver
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.findViewTreeLifecycleOwner
 import androidx.lifecycle.lifecycleScope
 import androidx.recyclerview.widget.RecyclerView
 import com.yelp.android.bento.componentcontrollers.RecyclerViewComponentController.constructViewHolder
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentLinkedDeque
 import java.util.concurrent.Executors
+import kotlin.coroutines.CoroutineContext
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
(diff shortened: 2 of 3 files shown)


redwarp (approved):

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -343,6 +343,9 @@ public void clear() {
         mComponentGroup.clear();
         removeVisibilityListeners();
         addVisibilityListeners();
+        if (mAsyncInflationEnabled) {
+            mAsyncInflationBridge.cancelAllInflationJobs();
```
redwarp, line 347: ++

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -146,4 +150,30 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+            owner.lifecycle.addObserver(object : LifecycleObserver {
+
+                @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
+                fun cleanUp() {
+                    clearResources()
+                    owner.lifecycle.removeObserver(this)
+                }
+            })
+        }
+    }
```
redwarp: I don't think you need to wrap the addObserver in a coroutine, can simply do

```
private fun findAndAttachToLifecycleOwner() {
    var owner = recyclerView.findViewTreeLifecycleOwner()
    if (owner == null) {
        if (recyclerView.context is LifecycleOwner) {
            owner = recyclerView.context as LifecycleOwner
        }
    }
    owner?.lifecycle?.addObserver(object : LifecycleObserver {
        @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
        fun cleanUp() {
            clearResources()
            owner.lifecycle.removeObserver(this)
        }
    })
}
```
Adding an observer is async by definition right, plus the findAndAttachToLifecycleOwner() would only be called from the main thread, as the init would happen on the main thread.

At the very least, a `Dispatchers.Main.immediate` should be used instead of the `Dispatchers.Main`?

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -146,4 +150,30 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+
+    private fun clearResources() {
+        coroutineContext.cancelChildren()
```
redwarp, line 174: Neat!

redwarp (approved):
Fix and ship!

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: @redwarp If there is a RecyclerView in a ViewHolder then I think it will still be created on the background thread.

redwarp: Ah, good point! Let's go for Dispatchers.Main.immediate then!

martinpdd8: Sounds good!

martinpdd8:
@redwarp I think this is ready to be merged now.

PR #85, opened 2021-09-13
Smarter async view inflation.

martinpdd8:
This implementation for async view inflation tracks how many view holders and views were actually needed by a page in a cache, and then uses it to inflate the number of view holders when the first component is added the next time the page is created.

This has some benefits:

- We inflate less views in total and have less unused views when someone scrolls to the bottom. Therefore, it uses less memory overall.
- We get a higher hit % from the cache of pre-inflated views so the ui will be smoother. Some pages that add components without data were getting a lot of misses with the other implementation.

And some drawbacks:
- Navigating to a page one time bypasses async view inflation completely.
- This implementation blocks the ui waiting for above the fold components, although it's lightning fast, the height of components impacts how many are actually above the fold or not from page to page.

Note: I deleted the old implementation here but keeping both and allowing each RecyclerViewComponentController to specify a different async inflation strategy could be a worthwhile.

### DIFF
--- a/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/RecyclerViewActivity.kt
+++ b/bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/RecyclerViewActivity.kt
@@ -11,6 +11,7 @@ import com.yelp.android.bento.components.ListComponent
 import com.yelp.android.bento.components.NestedComponent
 import com.yelp.android.bento.components.NestedViewModel
 import com.yelp.android.bento.components.SimpleComponent
+import com.yelp.android.bento.core.AsyncInflationStrategy.DEFAULT
 import com.yelp.android.bento.core.Component
 import com.yelp.android.bento.core.ComponentController
 import com.yelp.android.bentosampleapp.components.AnimatedComponentExampleViewHolder
@@ -29,7 +30,11 @@ import kotlinx.android.synthetic.main.activity_recycler_view.*
 class RecyclerViewActivity : AppCompatActivity() {
 
     private val componentController by lazy {
-        RecyclerViewComponentController(recyclerView)
+        RecyclerViewComponentController(recyclerView).apply {
+            setAsyncCacheKey("RecyclerViewActivity")
+            setAsyncStrategy(DEFAULT)
+            setNumberAboveTheFoldViewHolders(3)
+        }
     }
     private lateinit var componentToScrollTo: Component
 
--- a/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
+++ b/bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
@@ -1,5 +1,6 @@
 package com.yelp.android.bento.componentcontrollers;
 
+import android.util.Log;
 import android.view.View;
 import android.view.ViewGroup;
 import androidx.annotation.NonNull;
@@ -13,6 +14,7 @@ import androidx.recyclerview.widget.RecyclerView.Orientation;
 import androidx.recyclerview.widget.RecyclerView.ViewHolder;
 import com.google.common.collect.HashBiMap;
 import com.yelp.android.bento.core.AsyncInflationBridge;
+import com.yelp.android.bento.core.AsyncInflationStrategy;
 import com.yelp.android.bento.core.BentoLayoutManager;
 import com.yelp.android.bento.core.Component;
 import com.yelp.android.bento.core.Component.ComponentDataObserver;
@@ -25,6 +27,7 @@ import com.yelp.android.bento.core.ComponentVisibilityListener;
 import com.yelp.android.bento.core.ComponentVisibilityListener.LayoutManagerHelper;
 import com.yelp.android.bento.core.ListItemTouchCallback;
 import com.yelp.android.bento.core.OnItemMovedPositionListener;
+import com.yelp.android.bento.core.SmartAsyncInflationCache;
 import com.yelp.android.bento.utils.AccordionList.Range;
 import com.yelp.android.bento.utils.AccordionList.RangedValue;
 import com.yelp.android.bento.utils.BentoSettings;
@@ -61,6 +64,9 @@ public class RecyclerViewComponentController
     @RecyclerView.Orientation private int mOrientation;
 
     private final boolean mAsyncInflationEnabled;
+    // This will be used to track how many of each view holder was needed when async inflation is
+    // enabled.
+    private String mAsyncCacheKey = null;
 
     /**
      * Creates a new {@link RecyclerViewComponentController} and automatically attaches itself to
@@ -112,6 +118,7 @@ public class RecyclerViewComponentController
         mAsyncInflationEnabled = asyncInflationEnabled;
         if (asyncInflationEnabled) {
             mAsyncInflationBridge = new AsyncInflationBridge(recyclerView);
+            mAsyncCacheKey = mAsyncInflationBridge.getAsyncCacheKey();
         }
         mOrientation = orientation;
         mRecyclerViewAdapter = new RecyclerViewAdapter();
@@ -408,6 +415,46 @@ public class RecyclerViewComponentController
         mLayoutManager.setScrollEnabled(isScrollable);
     }
 
+    /**
+     * Allows overwriting of the default cache key used in {@link SmartAsyncInflationCache}.
+     *
+     * @param cacheKey Cache key which should be unique per controller / RecyclerView.
+     */
+    public RecyclerViewComponentController setAsyncCacheKey(@NonNull String cacheKey) {
+        if (mAsyncInflationEnabled) {
+            mAsyncCacheKey = cacheKey;
+            this.mAsyncInflationBridge.setAsyncCacheKey(cacheKey);
+        }
+        return this;
+    }
+
+    /**
+     * Sets the {@link AsyncInflationStrategy} to use for this controller if async inflation is
+     * enabled.
+     *
+     * @param strategy The strategy to use.
+     */
+    public RecyclerViewComponentController setAsyncStrategy(
+            @NonNull AsyncInflationStrategy strategy) {
+        if (mAsyncInflationEnabled) {
+            mAsyncInflationBridge.setStrategy(strategy);
+        }
+        return this;
+    }
+
+    /**
+     * Sets the number of view holdrs that are visible in the RecyclerView before scrolling. If this
+     * number is too high, it may cause UI issues. It's only used when async inflation is enabled.
+     *
+     * @param number The number of view holders visible before scrolling or "above the fold."
+     */
+    public RecyclerViewComponentController setNumberAboveTheFoldViewHolders(int number) {
+        if (mAsyncInflationEnabled) {
+            mAsyncInflationBridge.setNumberOfAboveTheFoldViewHolders(number);
+        }
+        return this;
+    }
+
     public void onRecyclerViewDetachedFromWindow() {
         mComponentVisibilityListener.onComponentGroupVisibilityChanged(false);
     }
@@ -588,6 +635,15 @@ public class RecyclerViewComponentController
      */
     private final class RecyclerViewAdapter extends RecyclerView.Adapter<ViewHolderWrapper>
             implements Sequenceable {
+        private float preInflatedViewUsed = 0;
+        private float nonPreInflatedViewUsed = 0;
+
+        // If we should track how many view holders were needed in SmartAsyncInflationCache.
+        private boolean shouldTrack = false;
+
+        private final Set<Class<? extends ComponentViewHolder>> missingViewHoldersSet =
+                new HashSet<>();
+
         @NonNull
         @SuppressWarnings("unchecked") // Unchecked Component generics.
         @Override
@@ -605,8 +661,45 @@ public class RecyclerViewComponentController
                 return new ViewHolderWrapper(viewHolder.inflate(parent), viewHolder);
             }
             View view = mAsyncInflationBridge.getView(viewHolder);
+            if (mAsyncCacheKey != null) {
+                if (!SmartAsyncInflationCache.INSTANCE.containsKey(mAsyncCacheKey)) {
+                    // Only track the first one so we don't end up with tons of views for
+                    // each view holder.
+                    shouldTrack = true;
+                }
+                if (shouldTrack || missingViewHoldersSet.contains(viewHolderType)) {
+                    if (BentoSettings.getLoggingEnabled()
+                            && missingViewHoldersSet.contains(viewHolderType)) {
+                        Log.i(BentoSettings.BENTO_TAG, "updating record for: " + viewHolderType);
+                    }
+                    SmartAsyncInflationCache.INSTANCE.incrementForViewHolder(
+                            mAsyncCacheKey, viewHolderType);
+                }
+                // Some pages have different components every time so some view holders might not
+                // have a cache hit. This attempts to track that 1 view holder was needed in such
+                // cases.
+                boolean wasMissing =
+                        SmartAsyncInflationCache.INSTANCE.incrementForViewHolderIfMissing(
+                                mAsyncCacheKey, viewHolderType);
+                if (wasMissing) {
+                    missingViewHoldersSet.add(viewHolderType);
+                }
+            }
+
             if (view == null) {
+                if (BentoSettings.getLoggingEnabled()) {
+                    nonPreInflatedViewUsed += 1.0f;
+                    Log.i(
+                            BentoSettings.BENTO_TAG,
+                            "% pre-inflated: " + getPercentage() + " FOR: " + parent);
+                }
                 return new ViewHolderWrapper(viewHolder.inflate(parent), viewHolder);
+            }
+            if (BentoSettings.getLoggingEnabled()) {
+                preInflatedViewUsed += 1.0f;
+                Log.i(
+                        BentoSettings.BENTO_TAG,
+                        "% pre-inflated: " + getPercentage() + " FOR: " + parent);
             }
             return new ViewHolderWrapper(view, viewHolder);
         }
@@ -649,6 +742,12 @@ public class RecyclerViewComponentController
         @Override
         public Sequence<Object> asItemSequence() {
             return ComponentControllerX.asItemSequence(RecyclerViewComponentController.this);
+        }
+
+        // Get the percentage of pre-inflated views used vs normal views.
+        private float getPercentage() {
+            if (preInflatedViewUsed == 0f && nonPreInflatedViewUsed == 0f) return 0f;
+            return preInflatedViewUsed / (preInflatedViewUsed + nonPreInflatedViewUsed) * 100;
         }
     }
 
--- a/bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
+++ b/bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
@@ -9,12 +9,17 @@ import androidx.lifecycle.findViewTreeLifecycleOwner
 import androidx.lifecycle.lifecycleScope
 import androidx.recyclerview.widget.RecyclerView
 import com.yelp.android.bento.componentcontrollers.RecyclerViewComponentController.constructViewHolder
+import com.yelp.android.bento.core.AsyncInflationStrategy.BEST_GUESS
+import com.yelp.android.bento.core.AsyncInflationStrategy.DEFAULT
+import com.yelp.android.bento.core.AsyncInflationStrategy.SMART
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentLinkedDeque
 import java.util.concurrent.Executors
+import java.util.concurrent.atomic.AtomicInteger
 import kotlin.coroutines.CoroutineContext
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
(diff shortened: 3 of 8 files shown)


bento/src/main/java/com/yelp/android/bento/core/ViewPrecognitionCache.kt
```diff
@@ -0,0 +1,26 @@
+package com.yelp.android.bento.core
+
+// Cache of how many of each view holder type's were actually inflated.
+object ViewPrecognitionCache {
+    // This is a map of a name of a screen, to map of ViewHolder type's to how many view holder
+    // type's were requested.
+    @JvmStatic val requiredViewHolders = mutableMapOf<String, MutableMap<Any, Int?>?>()
+
+    // TODO make neater
+    fun incrementForViewHolder(name: String, viewHolderType: Any) {
```
redwarp: Instead of a name, would a hashCode of the recyclerview be sufficient instead? So it could be automatically applied to all? (Or, if no name, use a hashcode).

The idea being that what causes problem is often when we add and remove to many views to one single recyclerview.

It would not prevent memory bloating if we navigate between lots of recycler views though, but could be a good middle ground?

bento/src/main/java/com/yelp/android/bento/core/SmartAsyncInflationCache.kt
```diff
@@ -0,0 +1,26 @@
...
+ * a screen, to map of ViewHolder type's to how many view holder type's were requested.
+ */
+object SmartAsyncInflationCache : LinkedHashMap<String, MutableMap<Any, Int?>>() {
+
+    fun incrementForViewHolder(name: String, viewHolderType: Any) {
+        val existingMap = this[name] ?: mutableMapOf()
+        val existingCount = (existingMap[viewHolderType] ?: 0) + 1
+        existingMap[viewHolderType] = existingCount
+        this[name] = existingMap
+    }
```
redwarp: Nit: could be
```
fun incrementForViewHolder(name: String, viewHolderType: Any) {
    val existingMap = this.getOrPut(name) { mutableMapOf() }
    val existingCount = (existingMap[viewHolderType] ?: 0) + 1
    existingMap[viewHolderType] = existingCount
}
```

bento/src/main/java/com/yelp/android/bento/core/ViewPrecognitionCache.kt
martinpdd8: We talked on Slack about why the hashCode won't work.

bento-sample-app/src/main/java/com/yelp/android/bentosampleapp/RecyclerViewActivity.kt
```diff
@@ -29,7 +30,11 @@ import kotlinx.android.synthetic.main.activity_recycler_view.*
 class RecyclerViewActivity : AppCompatActivity() {
 
     private val componentController by lazy {
-        RecyclerViewComponentController(recyclerView)
+        RecyclerViewComponentController(recyclerView).apply {
+            setAsyncName("asdf")
```
dwaxemberg: Maybe make this name something believable? Like `Main Screen Fragment` or something?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -258,7 +264,7 @@ public RecyclerViewComponentController addComponent(
             mAsyncInflationBridge.asyncInflateViewsForComponent(
                     component,
                     () -> {
-                        addComponentInternal(index, component);
+                        addComponentInternal(component);
```
dwaxemberg: Should we keep the index param here?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -408,6 +414,29 @@ public void setScrollable(boolean isScrollable) {
         mLayoutManager.setScrollEnabled(isScrollable);
     }
 
+    public RecyclerViewComponentController setAsyncName(@NonNull String name) {
```
dwaxemberg: Maybe add javadoc?

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -588,6 +617,20 @@ private void setupComponentSpans() {
...
+        // Get the percentage of pre-inflated views used vs normal views.
+        private float getPercentage() {
+            if (preInflatedViewUsed == 0f && nonPreInflatedViewUsed == 0f) return 0f;
+            return preInflatedViewUsed / (preInflatedViewUsed + nonPreInflatedViewUsed) * 100;
+        }
+
+        // If we should track how many view holders were needed in ViewPrecognitionCache.
+        private boolean shouldTrack = false;
+
+        private Set<Class<? extends ComponentViewHolder>> missingViewHoldersSet = new HashSet<>();
```
dwaxemberg: nit: move the method below the fields

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +114,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+            // If there's more than 10, chances are they'll be recycled and more will go unused.
+            // Not gonna lie. It's a guess and could be improved.
+            if (component.count > 10) {
+                numberOfViewsToInflate = component.count / 2
+            }
+            // If this amount would tip us over the MAX views value, only inflate enough
+            // to fill up to the max.
+            if (numberOfViewsToInflate + inflatedViewTracker.get() > MAX_VIEWS) {
+                numberOfViewsToInflate = (MAX_VIEWS - inflatedViewTracker.get())
+                        .coerceAtMost(numberOfViewsToInflate)
```
dwaxemberg: This coerceAtMost seems weird since we're assigning it back to the value. Shouldn't we just do:
`numberOfViewsToInflate.coerceAtMost(MAX_VIEW - inflatedViewTracker.get())`?

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +114,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+     * its count is non-zero and if doing so doesn't inflate more
+     * than [MAX_VIEWS].
+     */
+    private fun asyncInflateViewsBestGuess(component: Component, addComponentCallback: () -> Unit) {
+        inflationJobs[component] = launch(defaultBridgeDispatcher) {
+            var numberOfViewsToInflate = component.count
+            // If there's more than 10, chances are they'll be recycled and more will go unused.
+            // Not gonna lie. It's a guess and could be improved.
+            if (component.count > 10) {
+                numberOfViewsToInflate = component.count / 2
```
dwaxemberg: Why divide by 2? why not cap at some arbitrary number (like 10)?

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +114,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+     * The "BEST_GUESS" strategy for async view inflation. Inflates views for [component] if
+     * its count is non-zero and if doing so doesn't inflate more
+     * than [MAX_VIEWS].
+     */
+    private fun asyncInflateViewsBestGuess(component: Component, addComponentCallback: () -> Unit) {
+        inflationJobs[component] = launch(defaultBridgeDispatcher) {
+            var numberOfViewsToInflate = component.count
+            // If there's more than 10, chances are they'll be recycled and more will go unused.
+            // Not gonna lie. It's a guess and could be improved.
+            if (component.count > 10) {
```
dwaxemberg: nit: we should use `numberOfViewsToInflate` instead of calling `component.count` again since that's a function call which _could_ return a different value the second time. #concurrentprogramming

bento/src/main/java/com/yelp/android/bento/core/SmartAsyncInflationCache.kt
```diff
@@ -0,0 +1,25 @@
+package com.yelp.android.bento.core
+
+/**
+ * Cache of how many of each view holder type's were actually inflated. This is a map of a name of
+ * a screen, to map of ViewHolder type's to how many view holder type's were requested.
+ */
+object SmartAsyncInflationCache : LinkedHashMap<String, MutableMap<Any, Int?>>() {
```
dwaxemberg, line 9: Should we be worried about this cache taking up all the memory? Unless I missed something, we never clear it right? So we will track every viewholder ever in the app right?

dwaxemberg (approved):
overall this is looking good!

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -53,7 +62,19 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
     private val viewMap = ConcurrentHashMap<ComponentViewHolder<*, *>, View?>()
     private val inflationJobs = ConcurrentHashMap<Component, Job>()
 
+    private var belowTheFoldTriggered = false
+    private var aboveTheFoldTriggered = false
+    private var smartAsyncCacheEmptyOnStart = false
```
redwarp: I would add some dokka doc above each of this variable to. explain how they are used

redwarp: `private val smartAsyncCacheEmptyOnStart = !SmartAsyncInflationCache.containsKey(name)`

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
redwarp: I would extract 10 to a. const

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +114,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
      * Cancels [component]'s inflation job if the component was removed.
      */
     fun trackComponentRemoval(component: Component) {
-        inflationJobs[component]?.cancel()
+        inflationJobs.remove(component)?.cancel()
```
redwarp, line 129: ++

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +114,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+                withContext(Dispatchers.Main) {
+                    executeAddComponentCallback(addComponentCallback)
+                }
+            }
+            // Outside the lock. The second time it's called, it should inflate the below the fold
+            // view holders without blocking any more addComponent calls.
+            if (!belowTheFoldTriggered && !smartAsyncCacheEmptyOnStart &&
+                    SmartAsyncInflationCache.containsKey(name)) {
+                asyncInflateViewsIfPossible(aboveFold, false)
+            }
```
redwarp: This part confuses me, the fact that we can call async inflate in some case after the callback has been executed.
I will need to re-read that tomorrow.

redwarp (commented):
Sorry, I'll have to re-read that again and run through the code with a few break point, there are still some areas that I can't wrap my head around.

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
martinpdd8: @dwaxemberg I ended up renaming it too 👍

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: That's a great question. I really like that idea though. Just a blanket rule to cap at 10 per component if it's >= 10.

bento/src/main/java/com/yelp/android/bento/core/SmartAsyncInflationCache.kt
martinpdd8: Yeah. I didn't think Class's and Int's would take up much memory, but I can add a cap of like 30 or 40?

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +122,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+            // Not gonna lie. It's a guess and could be improved.
+            if (numberOfViewsToInflate >= VIEWS_PER_COMPONENT_THRESHOLD) {
+                numberOfViewsToInflate = VIEWS_PER_COMPONENT_THRESHOLD
+            }
+            // If this amount would tip us over the MAX views value, only inflate enough
+            // to fill up to the max.
+            if (numberOfViewsToInflate + inflatedViewTracker.get() > MAX_VIEWS) {
+                numberOfViewsToInflate = numberOfViewsToInflate
+                        .coerceAtMost(MAX_VIEWS - inflatedViewTracker.get())
+            }
```
redwarp: You can simplify that by simple `numberOfViewsToInflate = numberOfViewsToInflate .coerceAtMost(MAX_VIEWS - inflatedViewTracker.get())` : the content of the coerce at most method already does the > check.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +122,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+     * than [MAX_VIEWS].
+     */
+    private fun asyncInflateViewsBestGuess(component: Component, addComponentCallback: () -> Unit) {
+        inflationJobs[component] = launch(defaultBridgeDispatcher) {
+            var numberOfViewsToInflate = component.count
+            // If there's more than 10, chances are they'll be recycled and more will go unused.
+            // Not gonna lie. It's a guess and could be improved.
+            if (numberOfViewsToInflate >= VIEWS_PER_COMPONENT_THRESHOLD) {
+                numberOfViewsToInflate = VIEWS_PER_COMPONENT_THRESHOLD
+            }
```
redwarp: This probably could also be a coerceAtMost

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +122,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+            // If this amount would tip us over the MAX views value, only inflate enough
+            // to fill up to the max.
+            if (numberOfViewsToInflate + inflatedViewTracker.get() > MAX_VIEWS) {
+                numberOfViewsToInflate = numberOfViewsToInflate
+                        .coerceAtMost(MAX_VIEWS - inflatedViewTracker.get())
+            }
+            // Skip the below coroutine if there are either 0 views to inflate, or if it'd tip us
+            // over the max number of inflated views.
+            if (numberOfViewsToInflate > MAX_VIEWS || numberOfViewsToInflate == 0) {
+                lock.withLock {
```
redwarp, line 234: ++

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -53,6 +63,25 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+    // Keeps track of whether or not we inflated the below the fold set of views yet.
+    private var belowTheFoldTriggered = false
+
+    // Keeps track of whether or not we inflated the above the fold set of views yet.
+    private var aboveTheFoldTriggered = false
+
+    // Tracks how many views have been inflated.
+    private val inflatedViewTracker = AtomicInteger(0)
+
+    var asyncCacheKey: String = recyclerView.id.toString()
```
redwarp, line 75: Not sure if using the recyclerViews's id as default is a good idea, as a lot of layouts tend to use the same id for their recycler views (Something like R.id.list, or R.id.recycler_view)

redwarp: Changing the async cache key does not reset the `smartAsyncCacheEmptyOnStart` val.
I think it should be the case, otherwise we inflate views based on wrong assumptions.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +122,150 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+                }
+                withContext(Dispatchers.Main) {
+                    executeAddComponentCallback(addComponentCallback)
+                }
+            }
+            // Outside the lock. The second time it's called, it should inflate the below the fold
+            // view holders without blocking any more addComponent calls.
+            if (!belowTheFoldTriggered && !smartAsyncCacheEmptyOnStart &&
+                    SmartAsyncInflationCache.containsKey(asyncCacheKey)) {
+                asyncInflateViewsIfPossible(aboveFold, false)
```
redwarp: I would prepend here with `belowTheFoldTriggered = true` and remove
```
if (!isAboveTheFoldCall) {
    belowTheFoldTriggered = true
}
```
from the `asyncInflateViewsIfPossible` method to make it simpler

bento/src/main/java/com/yelp/android/bento/componentcontrollers/RecyclerViewComponentController.java
```diff
@@ -605,9 +661,45 @@ public ViewHolderWrapper onCreateViewHolder(@NonNull ViewGroup parent, int viewT
...
             View view = mAsyncInflationBridge.getView(viewHolder);
+            if (mAsyncCacheKey != null) {
+                if (!SmartAsyncInflationCache.INSTANCE.containsKey(mAsyncCacheKey)) {
+                    // Only track the first one so we don't end up with tons of views for
+                    // each view holder.
+                    shouldTrack = true;
+                }
+                if (shouldTrack || missingViewHoldersSet.contains(viewHolderType)) {
+                    if (missingViewHoldersSet.contains(viewHolderType)) {
+                        Log.i(BentoSettings.BENTO_TAG, "updating record for: " + viewHolderType);
```
redwarp, line 673: Probably also wrap in the `if (BentoSettings.getLoggingEnabled())`

bento/src/main/java/com/yelp/android/bento/core/SmartAsyncInflationCache.kt
```diff
@@ -0,0 +1,31 @@
+package com.yelp.android.bento.core
+
+private const val CACHE_SIZE_LIMIT = 30
```
redwarp, line 3: There seems to be a mechanism that can be used by LinkedHashMap to limit the size to x. In this exemple: https://stackoverflow.com/a/11469731/870751 they implement the removeEldestEntry and it seems to automatically remove the oldest one past the size limit.
It would be useful here!

martinpdd8: That's great! Thanks.

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +126,140 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+
+    @Suppress("UNCHECKED_CAST")
+    private suspend fun asyncInflateViewsIfPossible(isAboveTheFoldCall: Boolean) {
+        if (asyncCacheKey.isEmpty()) return // There's no work to do.
+
+        val mapToPrepare = SmartAsyncInflationCache[asyncCacheKey] ?: mutableMapOf()
+        if (mapToPrepare.isEmpty()) return // There's no work to do.
+
+        var count = numberOfAboveTheFoldViewHolders
+        val aboveFoldKeys = mapToPrepare.keys.takeWhile { count--; count > -1 }
```
redwarp: I think you can simplify this by `val aboveFoldKeys = mapToPrepare.keys.take(numberOfAboveTheFoldViewHolders)`

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
```diff
@@ -116,13 +126,140 @@ internal class AsyncInflationBridge @JvmOverloads constructor(
...
+            numberOfViewsToInflate.coerceAtMost(VIEWS_PER_COMPONENT_THRESHOLD)
+
+            // If this amount would tip us over the MAX views value, only inflate enough
+            // to fill up to the max.
+            numberOfViewsToInflate = numberOfViewsToInflate
+                    .coerceAtMost(MAX_VIEWS - inflatedViewTracker.get())
+
+            // Skip the below coroutine if there are either 0 views to inflate, or if it'd tip us
+            // over the max number of inflated views.
+            if (numberOfViewsToInflate > MAX_VIEWS || numberOfViewsToInflate == 0) {
```
redwarp: As it's coerced to be at most MAX_VIEWS minus something, I think the condition `numberOfViewsToInflate > MAX_VIEWS` is never true.

redwarp (approved):

bento/src/main/java/com/yelp/android/bento/core/AsyncInflationBridge.kt
martinpdd8: @redwarp I removed it from the condition.

martinpdd8 (commented):
@redwarp @dwaxemberg I think this is ready to merge now 🎉

PR #87, opened 2022-03-23
Added jetpack compose support. Bumped Kotlin and Gradle.

martinpdd8:
### Jetpack Compose Support

This PR adds a new module called `bento-compose` which provides a ViewHolder that in which Jetpack Compose can be used. It also allows `RecyclerView` to be used directly in a `@Composable`.

This change also:
1. Bumps Kotlin to v1.6.10
2. Bumps Gradle to 7.4.2
3. Bumps targetSdkVersion and compileSdkVersion to 31
4. Bumps AGP to 7.1.2
5. Deletes the 3rd party maven-publish plugin as its officially supported by the Android Gradle Plugin now and wouldn't work anymore.

Note: I verified that `publishToMavenLocal` works but I didn't yet verify the other publish tasks still work...

PR #89, opened 2022-04-27
feat: Added jetpack compose support and updated build tools and libra…

martinpdd8:
…ries.

BREAKING CHANGE: Increasing the minSDK and targetSDK may mean this version is no longer compatible with some apps.

This change is just https://github.com/Yelp/bento/pull/87 again with a breaking change commit so we can actually release it.


YingChen0605 (approved):
LGTM!

PR #90, opened 2022-04-27
fix(publishing): Fixing publishing to Maven Central

dwaxemberg:
The signing task had been removed, so we weren't properly signing artifacts when publishing to maven. The task is now added back and working properly.


martinpdd8 (approved):
Thanks!

PR #94, opened 2023-03-20
Bump compileSdkVersion and  targetSdkVersion to 33, AGP 7.4.2, Gradle 7.5.1, Kotlin 1.8.10, Compose 1.4.0, Compose Kotlin Compiler 1.4.4, RecyclerView 1.3.0

YingChen0605:
### Bump libs to embrace better Compose performance since RecyclerView 1.3.0
According to [Compose official document](https://developer.android.com/jetpack/compose/migrate/other-considerations#compose-recyclerview), Composables in RecyclerView are performant since RecyclerView version 1.3.0-alpha02. 
And now [RecyclerView of 1.3.0](https://developer.android.com/jetpack/androidx/releases/recyclerview#recyclerview-1.3.0) is officially released.

In order to embrace the performance win, we bump following libs:

- compileSdkVersion to 33
> this requires changing onViewDetachedFromWindow and onViewAttachedToWindow's parameter View to non-null 

- targetSdkVersion to 33
- Android Gradle Plugin to 7.4.2
- Gradle to 7.5.1
- Compose to 1.4.0
- RecyclerView to 1.3.0
> Note 1:
 In this version, we don't need to set [ViewCompositionStrategy](https://developer.android.com/reference/kotlin/androidx/compose/ui/platform/ViewCompositionStrategy)  because by default when a ComposeView is used as an item in a RecyclerView, composables are no longer disposed, but rather re-used. So we remove the ViewCompositionStrategy set previously and remove onViewRecycled function with we use for disposing composables.
>   more in https://medium.com/androiddevelopers/jetpack-compose-interop-using-compose-in-a-recyclerview-569c7ec7a583

> Note 2:
RecyclerView1.3.0 requires COMPILE_SDK to be >=33

- Kotlin to 1.8.10
> Followed https://developer.android.com/topic/libraries/view-binding/migration to migrate from Kotlin synthetics to Jetpack view binding, which is required for Kotlin 1.8.10
- Compose Kotlin Compiler to 1.4.4
> Compose Kotlin Compiler and Kotlin version needs to be compatible


aanandshekharroy (approved):
ShipIt!

bento-compose/src/main/java/com/yelp/android/bento/compose/ComposeViewHolder.kt
```diff
@@ -23,9 +22,6 @@ abstract class ComposeViewHolder<P, T> : ComponentViewHolder<P, T>() {
 
     final override fun inflate(parent: ViewGroup): View {
         return ComposeView(parent.context).apply {
-            setViewCompositionStrategy(
-                ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed
-            )
             setContent {
                 state = remember { mutableStateOf(element) }
```
martinpdd8: It looks like in the medium blog post they use a different strategy to this:

```kotlin
    fun bind(input: String) {
        composeView.setContent {
            MdcTheme {
                Text(input)
            }
        }
    }
```
We should probably switch the implementation here to match. Thoughts?

martinpdd8 (approved):
Looks great! Thanks for doing this.

Kurt-Bonatz (approved):
Looks good to me! Thanks for doing this!

PR #96, opened 2023-09-12
Compose view holder fails to update correctly when the component state changes

abdallaadelessa:
### Bug:
It seems that the Compose view holder fails to update correctly when the component state changes. This issue mainly arises from the fact that `private val rowStates = mutableMapOf<Int, MutableState<T>>()` caches the old data. I'm not sure why we are caching the data because it doesn't seem necessary, considering that the data is already a parameter in method `override fun bind(presenter: P, element: T)`.

### Testing Scenario:
First Update:
Screenshot 2023-09-12 at 13 12 51
Second Update:
Screenshot 2023-09-12 at 13 12 59
Third Update:
Screenshot 2023-09-12 at 13 13 06

### Verification 
"ComposeViewHolderTest.`compose view holder handle updates correctly`"

### DIFF
--- a/bento-compose/src/main/java/com/yelp/android/bento/compose/ComposeViewHolder.kt
+++ b/bento-compose/src/main/java/com/yelp/android/bento/compose/ComposeViewHolder.kt
@@ -4,7 +4,9 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.ComposeView
 import com.yelp.android.bento.core.ComponentViewHolder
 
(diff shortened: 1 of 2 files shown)


bento-compose/src/main/java/com/yelp/android/bento/compose/ComposeViewHolder.kt
```diff
@@ -15,26 +17,27 @@ import com.yelp.android.bento.core.ComponentViewHolder
...
+            }
         }
         return composeView
     }
 
     override fun bind(presenter: P, element: T) {
-        this.presenter = presenter
-        this.element = element
-        composeView.setContent {
-            val state = rowStates.getOrPut(absolutePosition) {
```
abdallaadelessa, line 33: i am not sure whether the `BindView` method will be omitted when using `setContent` if the `presenter` and `element` instances remain unchanged.
 
 That's why i prefer to call method `setContent` just one time as we do in fragments and use `MutableState` to notify compose about the changes.
 
 `MutableState` won't notify the composable function if the new date is equal to the old one

bento-compose/src/main/java/com/yelp/android/bento/compose/ComposeViewHolder.kt
```diff
@@ -15,26 +17,27 @@ import com.yelp.android.bento.core.ComponentViewHolder
 abstract class ComposeViewHolder<P, T> : ComponentViewHolder<P, T>() {
 
     private lateinit var composeView: ComposeView
-    var presenter: P? = null
-    var element: T? = null
-    private val rowStates = mutableMapOf<Int, MutableState<T>>()
```
abdallaadelessa, line 20: ### BUG
It seems that this map caches the old value

YingChen0605 (approved):
Looks good to me! Thank you for fixing this!

PR #98, opened 2023-09-13
Bump AGP 8.0.2, Gradle 8.3, Kotlin 1.8.21, Prod & Test Libs to latest

abdallaadelessa:
### Fix a crash in the sample app
* ToggleScrollInRecyclerViewActivity

### Bump Libs
* CompileSdk -> 34
* AGP -> 8.0.2
* Gradle -> 8.3
* Kotlin -> 1.8.21
* Compose -> 1.5.0

### Fix android tests
* MainActivityTest
* RecyclerViewActivityTest


.github/workflows/pull_requests.yml
```diff
@@ -41,14 +53,21 @@ jobs:
           key: cache-gradle-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}
           restore-keys: |
             cache-gradle-
+      - name: Delete Maven Settings file
+        run: rm ~/.m2/settings.xml
```
abdallaadelessa, line 57: After installing  java 17
i faced a problem with the maven publish plugin `net.linguica.maven-settings`

it seems that we have some servers here `~/.m2/settings.xml` but we don’t have `~/.m2/settings-security.xml` (with master password) to encrypt the servers passwords

The gradle was crashing here
https://github.com/mark-vieira/gradle-maven-settings-plugin/blob/master/src/main/groovy/net/linguica/gradle/maven/settings/LocalMavenSettingsLoader.java#L89

My workaround is to just delete the `~/.m2/settings.xml` before triggering the build :smile:

abdallaadelessa: it might be because of `setup-java` github action which i am using to install java 17
https://github.com/actions/setup-java/tree/main#maven-options
it seems to generate `~/.m2/settings.xml` after installing java 17

abdallaadelessa: it should be fine for now i think as long as it doesn’t impact development or final publishing

buildSrc/src/main/java/com/yelp/gradle/bento/GlobalDependencies.kt
```diff
@@ -9,33 +9,37 @@ object Publishing {
...
 
 object Versions {
-    const val COMPILE_SDK = 33
+    const val COMPILE_SDK = 34
     const val TARGET_SDK = 33
     const val MIN_SDK = 21
 
+    // Java
+    const val SOURCE_COMPATIBILITY = 11
+    const val TARGET_COMPATIBILITY = 11
```
Kurt-Bonatz, line 18: Can you make sure to declare these updates in the change log: https://github.com/Yelp/bento/blob/master/CHANGELOG.md#-breaking-changes

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.3-all.zip
```
Kurt-Bonatz, line 3: You might need to run `./gradlew wrapper --gradle-version=8.3` once or twice more since I believe the jar and gradle wrapper executable have changed.

Kurt-Bonatz (approved):

gradle/wrapper/gradle-wrapper.properties
abdallaadelessa: Done!

buildSrc/src/main/java/com/yelp/gradle/bento/GlobalDependencies.kt
abdallaadelessa: i updated the commit message to include `!` in order to be marked as a breaking change in the change log 
https://www.conventionalcommits.org/en/v1.0.0/

dwaxemberg (approved):

